Issue #31, opened 2025-08-14
[JAVA] cannot get TestSuites


AcevedoR:
### Describe the issue

when searching all TestSuites with the JAVA SDK, it fails when Deserializing JSON response:

```
Exception when calling TestSuitesApi#searchTestSuites
Status code: 0
Reason: null
Response headers: null
ApiException{code=0, responseHeaders=null, responseBody='null'}
	at io.kestra.sdk.internal.ApiClient.invokeAPI(ApiClient.java:1044)
	at io.kestra.sdk.api.TestSuitesApi.searchTestSuites(TestSuitesApi.java:900)
	at io.kestra.sdk.api.TestSuitesApi.searchTestSuites(TestSuitesApi.java:830)
	at io.kestra.plugin.ee.tests.POCSDK.main(POCSDK.java:27)
Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot construct instance of `io.kestra.sdk.model.PropertyString` (although at least one Creator exists): no String-argument constructor/factory method to deserialize from String value ('dont send Slack message')
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 1504] (through reference chain: io.kestra.sdk.model.PagedResultsTestSuite["results"]->java.util.ArrayList[0]->io.kestra.sdk.model.TestSuite["testCases"]->java.util.ArrayList[0]->io.kestra.sdk.model.UnitTest["fixtures"]->io.kestra.sdk.model.Fixtures["tasks"]->java.util.ArrayList[0]->io.kestra.sdk.model.TaskFixture["description"])
	at com.fasterxml.jackson.databind.exc.MismatchedInputException.from(MismatchedInputException.java:63)
	at com.fasterxml.jackson.databind.DeserializationContext.reportInputMismatch(DeserializationContext.java:1754)

```

### Environment

- Kestra Version: develop

AcevedoR:
The SDK is generated with io.kestra.sdk.model.PropertyString that it cannot deserialize to string, we had the same issue with go

PR #32, opened 2025-08-14
fix(java): add type mappings for Property<...> to string

AcevedoR:
- fixes https://github.com/kestra-io/client-sdk/issues/31

the main change is actually in: [configurations/java-config.yml](https://github.com/kestra-io/client-sdk/pull/32/files#diff-b352aa239f69ec42cc2c4ca3f3719506d798f51cdab29508d9b2c5fa9acd5f7d) (like for GO, we map every Property <> to String)

then I just re generated the JAVA SDK for a 0.24.1 version


Skraye (approved):
lgtm!

PR #128, opened 2025-11-18
fix(java-sdk): fix followDependenciesExecution return type

ClemDoum:

---
The `/api/v1/{tenant}/executions/{executionId}/follow-dependencies` API routes returns a stream of [`ExecutionStatusEvent`](https://github.com/kestra-io/kestra/blob/a4ca3498f3a54d330cf05053b97d50154c7d7b8f/webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java#L2428) yet the client code deserialize the event stream as a stream of `Execution` which ends up in a stream of empty objects when calling `KestraClient().builder().executions().followDependenciesExecution(executionId, MAIN_TENANT, false, true)` (because property names don't match)

---
Before:
```java
KestraClient().builder().executions().followDependenciesExecution(executionId, MAIN_TENANT, false, true).toStream().toList()
```
return a list of empty `Execution` objects (null properties every everywhere

After:
```java
KestraClient().builder().executions().followDependenciesExecution(executionId, MAIN_TENANT, false, true).toStream().toList()
```
**should** return a list of valid `ExecutionStatusEvent` (I wasn't able to test properly as I wasn't able to setup unit test properly)

### DIFF
--- a/java/java-sdk/src/main/java/io/kestra/sdk/api/ExecutionsApi.java
+++ b/java/java-sdk/src/main/java/io/kestra/sdk/api/ExecutionsApi.java
@@ -27,12 +27,11 @@ import io.kestra.sdk.internal.BaseApi;
 import io.kestra.sdk.internal.Configuration;
 import io.kestra.sdk.internal.Pair;
 
-import io.kestra.sdk.model.BulkErrorResponse;
 import io.kestra.sdk.model.BulkResponse;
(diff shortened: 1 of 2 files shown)


Skraye:
@ClemDoum can you rebase your branch so I can merge it ? I dont have permission to commit on your branch

PR #139, opened 2025-12-11
fix: avoid double testing scheme

elevatebart:
- move all testing in a different package so as to actually test the library
- add typechecking to the tests
- fix all the errors found
- check if changes were committed to generator but the sdk changes were not committed

### DIFF
--- a/.github/workflows/javascript-sdk.yml
+++ b/.github/workflows/javascript-sdk.yml
@@ -83,15 +83,8 @@ jobs:
           echo $APPLICATION_SECRETS | base64 -d > ../test-utils/docker-setup/application-secrets.yml
           ./run-tests.sh
 
-      # Trusted publishing need npm 11.+
-      - name: Upgrade npm
-        if: startsWith(github.ref, 'refs/tags/v')
-        run: |
-          npm install -g npm@latest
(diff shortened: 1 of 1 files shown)


.github/workflows/javascript-sdk.yml
```diff
@@ -71,15 +73,40 @@ jobs:
...
+      - name: Get OIDC Token
+        if: startsWith(github.ref, 'refs/tags/v')
+        env:
+          AUDIENCE: "npm:registry.npmjs.org"
+        run: |
+          TOKEN_JSON=$(curl -s -H "Authorization: bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" "$ACTIONS_ID_TOKEN_REQUEST_URL&audience=$AUDIENCE")
+          ID_TOKEN=$(echo "$TOKEN_JSON" | jq -r .value)
+          echo "NODE_AUTH_TOKEN=$ID_TOKEN" >> "$GITHUB_ENV"
+
+      - name: Set version
```
Skraye: Tag looks like `v{version}-{language}` so I dont think it will works
You already fetch it from the package.json in `extract_version` so use it

PR #147, opened 2026-01-13
feat(java,javascript,python): regenerate, improve gen by removing noise

AcevedoR:
- some cleanup
- update generate sh to remove generated files, to prevent old files being kept after renamings
- regenerate java, javascript and python


Skraye (approved):
LGTM

PR #200, opened 2026-03-27
fix(js-sdk): add types to exports field for modern moduleResolution

xuxusheng:
## Problem

The `@kestra-io/kestra-sdk` package.json `exports` field is missing a `types` condition:

```json
"exports": {
  ".": "./dist/KestraClient.mjs"
}
```

When TypeScript uses `moduleResolution: "bundler"` (Vite, Next.js, etc.) or `"node16"` / `"nodenext"`, it resolves imports through the `exports` map. Since there's no `types` condition, TypeScript cannot find the type declarations, even though `dist/KestraClient.d.ts` exists and the top-level `types` field points to it.

The official test project uses the legacy `moduleResolution: "node"` which ignores `exports` entirely — that's why it wasn't caught.

## Fix

Add the `types` condition to the `exports` field:

```json
"exports": {
  ".": {
    "types": "./dist/KestraClient.d.ts",
    "default": "./dist/KestraClient.mjs"
  }
}
```

## Error before fix

```
error TS7016: Could not find a declaration file for module '@kestra-io/kestra-sdk'.
There are types at '.../KestraClient.d.ts', but this result could not be resolved
when respecting package.json "exports".
```


elevatebart:
The exports in this package are generated by tsdown. They will be working soon.

PR #245, opened 2026-05-27
tech: add SDK API coverage script

jymaire:
Computes how much of the Kestra HTTP API each hand-written SDK covers, deriving the API surface directly from the backend Micronaut controllers (kestra OSS + kestra-ee) rather than from the OpenAPI spec, which is being removed. Matches SDK methods to backend endpoints on (verb, normalized path).

Reports Python/Go/Java coverage; JS is generated from OpenAPI at build time (src/openapi is gitignored) so it has no source to measure.


jymaire:
moved to plugins-devtools repo

Issue #144, opened 2026-01-08
SDK example code is not working (Python, JS, etc.)


Pradumnasaraf:
The example code for the SDK (Python, JS, etc.) is not working. This is a known issue and has already been noticed by Yann and others involved in the SDK work. 

Creating this issue for the sake of the QA process. https://www.notion.so/kestra-io/SDK-2df36907f7b5809fb3c1c1550df56102?v=2df36907f7b5803a9f8e000c06287943&source=copy_link

jymaire:
Wait for SDK rework to be completed by @nkwiatkowski  before to start this issue

jymaire:
Examples to fix are the ones in docs : https://github.com/kestra-io/client-sdk/tree/main/python/python-sdk/docs

jymaire:
@Pradumnasaraf work done on the four SDK. Will be taken in next release

Pradumnasaraf:
Thank you, @jymaire 🙏

PR #266, opened 2026-06-05
fix(python): repair broken SDK examples (#144) and keep them in sync via tests + CI

jymaire:
Fixes the broken Python SDK examples reported in #144, and adds tooling so the
examples can't silently drift from the SDK again.

## Why
The Python API classes were hand-written in #237, but the README example and the
generated per-API `docs/*.md` still reflected the old OpenAPI-generated shape —
so essentially every documented Python example was broken (wrong accessor, wrong
argument order, calls to renamed/removed methods, invalid keyword args, missing
`pprint` import).

### 1. Repair the examples (`fix(python)` commit)
- **README_PYTHON_SDK.md**: correct `search_flows` (tenant-first, defined var).
- **docs/*.md (10 files)**:
  - accessor `kestra_client.FlowsApi.*` → `kestra_client.flows.*` on all ~176 calls
    (incl. `KVApi`→`kv`, `ServiceAccountApi`→`service_account`, `TestSuitesApi`→`test_suites`);
  - fix 37 mis-ordered calls to the tenant-first hand-written signatures;
  - drop nonexistent `q=` (search_groups/search_roles), fix `file_upload=`→`file_content=`;
  - add the missing `from pprint import pprint` to every example block;
  - rename 4 stale methods, remove 3 sections for methods that no longer exist.

### 2. Keep them in sync (`feat(python)` commit) — see `design/examples-as-source-of-truth.md`
**Curated README example → single source of truth**
- `testApis/basic_sdk_usage_example.py`: the search+create-flow example, exercised
  end-to-end against live Kestra by `test_basic_sdk_usage_example.py`.
- `test-utils/embed_snippets.py`: language-agnostic injector; `--check` fails CI on
  drift, `--write` fixes it. The README usage block is injected from the tested file.

**Generated reference docs → validated against the code**
- `scripts/validate_doc_examples.py`: static validator (AST, no live server) that checks
  every `docs/*.md` example against the hand-written signatures.
- `generate-sdks.sh`: dropped the dead doc post-processor (wrong path; mangled
  `ServiceAccountApi`) and gated generation on the validator instead.

**CI**: both static checks run early in `python-sdk.yml`, before the Docker test.

## Testing
- Static checks pass locally: validator, injector `--check`, `py_compile`, `bash -n`.
- The validator was confirmed to **fail** on a deliberately mis-ordered call.
- ⚠️ The new live example test (`test_basic_sdk_usage_example.py`) was **not** run against
  a Kestra container locally — opening as draft to verify on CI.

## Scope
Python only. JavaScript / Java already have tested examples that can be wired to the same
injector; Go needs one. Fully retiring the legacy Python OpenAPI generation path is a
larger follow-up (see design doc §7).

### DIFF
--- a/design/examples-as-source-of-truth.md
+++ b/design/examples-as-source-of-truth.md
@@ -1,6 +1,6 @@
 # Keeping SDK examples correct: examples as a single source of truth
 
-**Status:** Proposal / design — no implementation yet
+**Status:** Implemented for Python (this branch / PR #266); JavaScript, Java and Go pending (see §5)
 **Motivation:** [issue #144](https://github.com/kestra-io/client-sdk/issues/144) — "SDK example code is not working (Python, JS, etc.)"
 **Author:** (draft for review)
 
--- a/design/examples-as-source-of-truth.md
+++ b/design/examples-as-source-of-truth.md
@@ -1,0 +1,189 @@
+# Keeping SDK examples correct: examples as a single source of truth
+
+**Status:** Proposal / design — no implementation yet
+**Motivation:** [issue #144](https://github.com/kestra-io/client-sdk/issues/144) — "SDK example code is not working (Python, JS, etc.)"
+**Author:** (draft for review)
+
+## 1. Problem
+
+The SDK ships example code in two distinct surfaces, and **both drift out of sync with the actual SDK** because they are hand- or generator-copied snapshots that nothing verifies:
+
+1. **Curated "getting-started" examples** — the snippets in the root `README_*_SDK.md` files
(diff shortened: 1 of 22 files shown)


design/examples-as-source-of-truth.md
```diff
@@ -0,0 +1,189 @@
+# Keeping SDK examples correct: examples as a single source of truth
+
+**Status:** Proposal / design — no implementation yet
```
fdelbrayelle: If this PR respect this decision then it's supposed to be already approved?

fdelbrayelle (approved):

PR #278, opened 2026-06-15
release(java): Java SDK 1.0.12

jymaire:
Java-only release of the Java SDK as **1.0.12** onto the `releases/v1.3.x` line.

## What this does
- Syncs the entire `java/` tree up to `main`'s state (9 Java commits' worth of changes).
- Bumps `java/java-sdk/gradle.properties` → `version=1.0.12`.
- **Scoped to `java/**` only** — `go-sdk/`, `python/`, `javascript/`, and `test-utils/` are untouched, so merging triggers **only** the Java SDK release workflow (the others are path-gated).

## On merge
Pushing to `releases/v1.3.x` runs the test job, then `publish` → Maven Central (`io.kestra:java-sdk:1.0.12`), a GitHub release with jars, and a Slack notification.

## Notes / risks
- The release branch carries an **older `java-sdk.yml`** (no `EmbedSnippets` README-sync step), so the missing `test-utils/EmbedSnippets.java` is not a problem here.
- Synced tests from `main` may reference newer `test-utils/` fixtures absent on `releases/v1.3.x`; if the test job fails on that, publish via `workflow_dispatch` with `skip-test: true`.


java/java-sdk/gradle.properties
```diff
@@ -1 +1 @@
-version=1.3
+version=1.0.12
```
jymaire, line 1: nope, to check

PR #286, opened 2026-06-17
ci(javascript): move npm release publish to tag-only release

jymaire:
Backport of #285 to the `releases/v1.3.x` line — moves the JavaScript SDK release to the tag-only model (terraform-provider-kestra / kestractl), mirroring the Python (#282) and Java (#284) backports.

## What changed
- **`javascript-sdk.yml`** → renamed *Javascript SDK release* → *Javascript SDK tests*; dropped the `publish-release` job. Keeps `test` + `publish-develop` (main-only dev snapshots).
- **`javascript-sdk-release.yml`** (new) → triggered on `v*-javascript` tags; resolves the version from the tag, regenerates the SDK at that version, publishes to npm, and moves the `vX.Y` dist-tag.

## Decisions (per #285)
- **Tag is authoritative for the version** — drops the old `npm view`-based patch auto-increment so the tag states the full version, consistent with the other repos.
- **`publish-develop` stays** in `javascript-sdk.yml` — it's a main-push snapshot pipeline, not a release.

## Notes
- Cherry-picked from #285. The cherry-pick conflicted only on the `publish-release` removal — `releases/v1.3.x` is behind `main` on `javascript-sdk.yml` (it still carries the manual `Get OIDC Token` step). I kept the release branch's own `test`/`publish-develop` jobs untouched and applied only the `publish-release` removal, so this PR pulls in no unrelated `main`-only changes.

### DIFF
--- a/.github/workflows/javascript-sdk-release.yml
+++ b/.github/workflows/javascript-sdk-release.yml
@@ -1,0 +1,60 @@
+name: Javascript SDK release
+
+# Tag-only release, aligned with terraform-provider-kestra / kestractl.
+# Push a tag like `v1.0.13-javascript` to publish that version to npm. The tag
+# is cut from an already-green release branch, so this workflow only builds and
+# publishes — tests run on push/PR via javascript-sdk.yml. The tag is the source
+# of truth for the version (no npm auto-increment). Per-commit develop snapshots
+# stay in javascript-sdk.yml (publish-develop job, main only).
+on:
+  push:
+    tags:
+      - "v*-javascript"
+
+jobs:
+  publish:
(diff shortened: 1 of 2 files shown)


Malaydewangan09 (approved):

elevatebart:
A few notes:

When creating a new tag, I have to do it in my terminal and push it. 
To increment the build, I have to already know what the previous published tag is.
I cannot create click a button to run an action to publish a version.

Also, since anybody can create and push a tag, there is no hard requirement for that version to pass tests.
This tag can be added to any branch, making it error prone.

If we already know about those caveats, then it's OK with me.

PR #288, opened 2026-06-17
feat(js-sdk): add useExpressionResolver composable for Pebble expression resolution

fdelbrayelle:
## Summary

- Ports [kestra-io/artifact-sdk#71](https://github.com/kestra-io/artifact-sdk/pull/71) to `@kestra-io/kestra-sdk`
- Plugin topology-details and modal components display raw task configuration that may contain Pebble expressions (`{{ vars.region }}`, `{{ namespace.x }}`, etc.) — these were shown verbatim instead of resolved values
- Layered resolution: client-side `vars.*` → `namespace.*` → optional server-side `evalExpression` when `executionId` is available

## Resolution layers

1. **`{{ vars.x }}`** — fetches flow `variables` map; works in editor without execution context
2. **`{{ namespace.x }}`** — fetches namespace variables (EE); requires optional `fetchNamespaceVars` fetcher
3. **`evalExpression`** — server-side fallback for remaining `{{ … }}` when `executionId` present; resolves `inputs`, `outputs`, `trigger`, `envs`, `globals`
4. **Unresolvable** — secrets/credentials returned as-is

## API

```ts
import { flow } from "@kestra-io/kestra-sdk/flows";
import { evalExpression } from "@kestra-io/kestra-sdk/executions";
import { useExpressionResolver } from "@kestra-io/kestra-sdk";

const { resolveSync, resolve } = useExpressionResolver(
  computed(() => props.namespace),
  computed(() => props.flowId),
  computed(() => props.execution?.id as string | undefined),
  { fetchFlow: flow, evalExpression },
);

// Sync (editor view, inside computed()):
const region = computed(() => resolveSync(props.taskRunner?.region));

// Async (post-execution, resolves inputs/outputs/trigger/envs):
const region = ref<string>();
watch(() => props.taskRunner?.region, async (v) => { region.value = await resolve(v); }, { immediate: true });
```

## Design notes

Fetchers are injected (not imported internally) so the composable stays decoupled from specific API modules. `vue` added as peer dependency (`^3.0.0`).

## Test plan

- [ ] Build passes (`npm run build` in `javascript/javascript-sdk`)
- [ ] Consumers import `useExpressionResolver` from `@kestra-io/kestra-sdk` and call with `fetchFlow` + optional `evalExpression`
- [ ] `resolveSync` resolves `{{ vars.x }}` against flow variables in editor view (no execution required)
- [ ] `resolve` falls back to `evalExpression` when `executionId` available and unresolved `{{ … }}` remain
- [ ] Unresolvable expressions (secrets, missing vars) returned as-is

### DIFF
--- a/javascript/javascript-sdk/package.json
+++ b/javascript/javascript-sdk/package.json
@@ -72,6 +72,9 @@
     "nprogress": "0.2.0",
     "axios": "1.16.1"
   },
+  "peerDependencies": {
+    "vue": "^3.0.0"
+  },
   "devDependencies": {
     "@hey-api/openapi-ts": "^0.97.3",
     "@types/node": "^25.9.1",
--- a/javascript/javascript-sdk/src/index.ts
+++ b/javascript/javascript-sdk/src/index.ts
@@ -5,6 +5,7 @@ import { client } from "./openapi/client.gen"
 import type { ClientOptions, Config } from "./openapi/client"
 
 export * from "./openapi/index"
+export * from "./useExpressionResolver"
 
 declare global {
     interface Window {
--- a/javascript/javascript-sdk/src/useExpressionResolver.ts
+++ b/javascript/javascript-sdk/src/useExpressionResolver.ts
@@ -1,0 +1,104 @@
+import { ref, watch, readonly } from "vue";
+import type { Ref } from "vue";
+
+function resolveStaticPatterns(
+    value: string,
+    contexts: {
+        vars?: Record<string, string>;
+        namespace?: Record<string, string>;
+    },
+): string {
+    return value.replace(/\{\{\s*([\w]+(?:\.[\w]+)*)\s*\}\}/g, (match, path: string) => {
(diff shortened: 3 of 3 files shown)


elevatebart (commented):
I do not understand the intention yet.
Can you add some tests to specify what you are trying to do? 

It feels like it does not belong here. But since I do not understand the purpose yet, I cannot say for sure.

javascript/javascript-sdk/package.json
```diff
@@ -72,6 +72,9 @@
     "nprogress": "0.2.0",
     "axios": "1.16.1"
   },
+  "peerDependencies": {
+    "vue": "^3.0.0"
+  },
```
elevatebart, line 77: this is a very very bad idea. All the consumers of the SDK will be prompted to add vue as a dependency which they most certainly will not use. What is they use React or Angular more popular and ubiquitous?

elevatebart (commented):
Meh

PR #299, opened 2026-06-25
feat(javascript): add expressions/render endpoint to EE spec and expose renderExpressions()

jymaire:
## Summary

- Adds the `POST /api/v1/{tenant}/expressions/render` endpoint (plus its
  `ExpressionController.RenderExpressionRequest` / `RenderedExpressions` schemas)
  to the EE OpenAPI spec (`kestra-ee.yml`).
- Adds a `./expressions` subpath export to the JavaScript SDK so openapi-ts
  generates a plain, framework-agnostic `renderExpressions()` function, importable
  from `@kestra-io/kestra-sdk/expressions`.

This replaces the earlier approach in this PR (a hand-written `useRenderedExpressions`
Vue composable). The SDK stays framework-agnostic: it ships only the generated
pass-through client, and consumers (EE plugin UIs) own their own framework glue.

## Endpoint behavior

The endpoint renders a list of Pebble expressions **for display purposes only**,
using a restricted engine:

| Expression | Result |
|---|---|
| `vars.*`, `flow.*`, `globals.*` | Resolved |
| `secret('KEY')` | `[secret: KEY]` — never the real value |
| `inputs.*`, `outputs.*`, `execution.*` | Resolved when `executionId` (or a flow context) is provided; raw otherwise |
| `env()`, `kv()`, non-deterministic / IO functions | Kept raw |
| Anything unresolvable | Kept raw (resolution is all-or-nothing per expression) |

Provide an `executionId` to resolve against an execution context, a `namespace` +
`flowId` (or a raw `flow` YAML source) to resolve against a flow context; otherwise
only globals are available.

## Usage

```ts
import { renderExpressions } from "@kestra-io/kestra-sdk/expressions";

const { rendered } = await renderExpressions({
    tenant,
    expressions: [task.region, task.bucket],
    executionId,
});
```

## Related

- kestra-io/kestra#16874
- Depends on the OSS endpoint implementation adding `POST /api/v1/{tenant}/expressions/render`


fdelbrayelle (approved):

elevatebart (changes requested):
This PR would make the SDK unusable in a React environment. 
Addin the vue lib as a peer dependency would make `npm install` fail in this context.

jymaire:
> This PR would make the SDK unusable in a React environment. Addin the vue lib as a peer dependency would make `npm install` fail in this context.

@elevatebart  ~~Do we have a React environment?~~ Is marking peer dep as optional like I did in the last [commit](https://github.com/kestra-io/client-sdk/pull/299/changes/878cef2ce0e32e7f8e123387613c18979bfc32db) good enough?

edit: remark from François : we could have users using SDK with React

kestra-ee.yml
```diff
@@ -35,6 +35,41 @@ tags:
 - name: Misc
   description: Misc api
 paths:
+  /api/v1/{tenant}/expressions/render:
```
jymaire, line 38: to replace once EE PR merged

jymaire:
Superseded by the auto-sync — `main` now carries both the generated `/expressions/render` spec block (via the kestra-ee@develop regen, PR #319) and the `./expressions` SDK export. Nothing left for this branch to add, so closing.

PR #300, opened 2026-06-26
feat(sdk): add asset lock/unlock endpoints across all SDKs

brian-mulier-p:
- Adds `assets().lockAsset(id, tenant, AssetLockRequest{ttl, executionId, flowId, flowNamespace, taskRunId})` and `assets().unlockAsset(id, tenant, executionId)` to the Java, Python, and Go SDKs, backing the EE asset-locking feature (kestra-io/kestra-ee#8667).
- JavaScript SDK already covers this via the OpenAPI generator; adds test coverage for lock/unlock (execution-owned lock release by matching executionId, mismatched executionId no-op).
- Consumed by plugin-kestra's Acquire/Release lock tasks (companion kestra-io/plugin-kestra PR).


javascript/test_javascript_sdk/vitest.config.ts
```diff
@@ -14,6 +14,7 @@ const fullPathSdkEntries = Object.fromEntries(
 const IGNORE_APIS = [
     "Ai",
     "Login",
+    "Policies",
```
jymaire: why this ignore?

python/python-sdk/docs/AssetsControllerApiAssetLock.md
```diff
@@ -0,0 +1,36 @@
...
+**flow_namespace** | **str** |  | [optional] 
+**namespace** | **str** |  | [optional] 
+**username** | **str** |  | [optional] 
+
+## Example
+
+```python
+from kestrapy.models.assets_controller_api_asset_lock import AssetsControllerApiAssetLock
+
+# TODO update the JSON string below
```
jymaire, line 22: TODO in this PR? (same for other MD file)

jymaire:
Reviewed across all four SDKs — models match `kestra-ee.yml` exactly, paths/verbs are right, and each SDK follows its file's existing conventions. Everything is purely additive, so per `AGENTS.md` there's no signature change to flag. A few things worth addressing:

### 1. The two new JS tests duplicate coverage already on `main`

`main`'s `AssetsApi.spec.ts` already has:

- `lockAsset + unlockAsset: manually locks then unlocks an asset`
- `lockAsset: the lock owner may still update the asset it holds the lock on`

The PR adds `lockAsset: acquires a manual user lock` and `unlockAsset: releases the lock so it can be re-acquired`, which are near-identical to the first one. The only net-new value is `expect(lock.ownerType).toBe("USER")` — probably better folded into the existing test than paid for with two extra round-trips against a live Kestra.

### 2. PR description is stale

The title and body say "Java SDK" + JS tests, but `b7e73fb` adds Go and Python — which is the bulk of the diff. The body also credits the PR with the "lock owner updating through its own lock" JS test, which already exists on `main`. Worth fixing since plugin-kestra consumers read this description.

### 3. `executionId` on `unlockAsset` is never exercised in any SDK

Go passes `nil`, Java `null`, Python and JS omit it. Per the spec that's the owner-checked release path:

> When set, the execution releases ONLY its own lock (owner-checked, no-op otherwise). When absent, a forced manual unlock removes the lock regardless of owner.

That's exactly the path plugin-kestra's Release task will take for execution locks, and it's currently untested. Same for `flowId` / `flowNamespace` / `taskRunId` on the lock request — no test sends them.

### 4. Nit: tests leave assets locked

`lockAsset_manual`, `lockAsset_sameOwnerReacquireExtends` and the new JS `acquires a manual user lock` all leave a `PT1H` lock behind. Asset IDs are randomized so there's no cross-test interference, and the surrounding tests don't clean up either — low priority, just accumulating state in the shared test tenant.

javascript/test_javascript_sdk/vitest.config.ts
brian-mulier-p: It was before your policy PR addition as a placeholder while this API gets handled

python/python-sdk/docs/AssetsControllerApiAssetLock.md
brian-mulier-p: It's never done anywhere else it seems? Another example but Claude says there are 345 other models concerned so I'd assume this PR is not the place to change that: https://github.com/kestra-io/client-sdk/blob/e16c86786840101ffd09df995bbadd3b1cd6642e/python/python-sdk/docs/AppsControllerApiApp.md

jymaire: ok, something to handle in another PR then

jymaire (approved):
LGTM — purely additive across all four SDKs, each following its own existing conventions:

- **Java** — `lockAsset`/`unlockAsset` use the same `invoke(...)` + `tenantPath(...)` + `queryParams(...)` shape as the neighbouring `deleteAsset`/`assetDependencies`.
- **Python** — `_json_request`/`_void_request` + `_build_query_params`, matching the rest of `assets_api.py`. Body serialization is fine: the request model exposes `to_dict()`, which `base_api.py` picks up.
- **Go** — `doJSON`/`doVoidJSON`, consistent with the other call sites.
- **JS** — generator output already covers this, so a test-only diff is correct.

Paths match the spec (`/api/v1/{tenant}/assets/{id}/lock`). Tests assert real values (`ownerType`, `lockedUntil` extension on same-owner re-acquire, owner-checked no-op on mismatched `executionId`) rather than just shape. No existing signature changed, so no `⚠️ Signature change` note is needed.

Both earlier review threads are settled — the `vitest.config.ts` ignore was reverted in e16c8678, and the docs `TODO` is generator-default across ~345 models, so it belongs in a separate PR.

⚠️ **Do not merge until #323 lands.** The red Java/Python checks are both the same unrelated failure — `DashboardsApiTest.exportChartToCsv_basic()` / `test_export_chart_to_csv_basic` (403) — fixed in #323. This branch was green on both on 2026-07-27; the regression came from the EE image, not from this PR. Rerun/rebase after #323 merges.

PR #315, opened 2026-07-02
ci(javascript): move npm release publish back into javascript-sdk.yml

jymaire:
## Problem

The JS SDK release publish fails with:

```
npm error code E404
npm error 404 Not Found - PUT https://registry.npmjs.org/@kestra-io%2fkestra-sdk
npm error 404 ... could not be found or you do not have permission to access it.
```

That E404-on-`PUT`-with-"no permission" is the signature of **npm Trusted Publishing (OIDC)** rejecting the token. npm binds a trusted publisher to a specific **repo + workflow filename**, registered against `javascript-sdk.yml`. Commit 48d3bca (#285) moved the publish step into a separate `javascript-sdk-release.yml`, so the OIDC token's `workflow_ref` claim no longer matched npm's config and the publish was denied.

Failing run: https://github.com/kestra-io/client-sdk/actions/runs/28595836505/job/84791356386

## Fix

Move the tag-only release job **back into `javascript-sdk.yml`** and delete the separate file:

- Add the `v*-javascript` tag pattern to `on.push`. `paths` filters apply to branch pushes only, not tags, so the tag trigger fires regardless of the paths list — no workflow split needed.
- Guard the `test` job with `if: ${{ !startsWith(github.ref, 'refs/tags/') }}` so it doesn't re-run on the release tag (cut from an already-green branch).
- `publish-develop` is already gated on `refs/heads/main`, so it never runs on a tag — untouched.
- Add `publish-release` job (the ex-release-file steps verbatim) gated on `startsWith(github.ref, 'refs/tags/v')` with `id-token: write` for OIDC.
- Delete `javascript-sdk-release.yml`.

## Before merge

Confirm npm's trusted-publisher config for `@kestra-io/kestra-sdk` names exactly `javascript-sdk.yml` for this repo.


brian-mulier-p (approved):
LGTM

PR #323, opened 2026-07-06
feat(dashboards): support CSV/ION format on chart export across all SDKs

brian-mulier-p:
- Renames the chart-export operations from CSV-only `*-to-csv` routes to a single `/export` route taking a `format` query param (`CSV` or `ION`)
- Updates the Java, Python, Go, and JavaScript SDKs plus their tests to match
- Breaking rename, not a deprecate-and-keep — old method names (`exportChartToCsv`, `exportDashboardChartDataToCSV`, etc.) are removed outright, matching the OSS route rename with no back-compat alias
- Hand-edits the EE OpenAPI spec (`kestra-ee.yml`) surgically — only the two export paths — rather than a full regen, to keep the diff scoped to this change
- Companion to kestra-io/kestra#17260 and kestra-io/plugin-kestra#157 (kestra-ee#9112); no SDK release until Kestra 2.0 ships — draft until then, tested locally via `publishToMavenLocal`/local builds only

### ⚠️ Signature change
- Java:   exportChartToCsv(tenant, request) → exportChart(tenant, request, format)
- Java:   exportDashboardChartDataToCSV(id, chartId, tenant, filters) → exportDashboardChart(id, chartId, tenant, filters, format)
- Python: export_chart_to_csv(tenant, request) → export_chart(tenant, request, format=None)
- Python: export_dashboard_chart_data_to_csv(id, chart_id, tenant, filters) → export_dashboard_chart(id, chart_id, tenant, filters, format=None)
- Go:     ExportChartToCsv(ctx, tenant, request) → ExportChart(ctx, tenant, request, format *string)
- Go:     ExportDashboardChartDataToCSV(ctx, id, chartId, tenant, filters) → ExportDashboardChart(ctx, id, chartId, tenant, filters, format *string)
- JS:     exportChartToCsv → exportChart, exportDashboardChartDataToCSV → exportDashboardChart (regenerated)

### DIFF
--- a/go-sdk/kestra_api_client/dashboards_api.go
+++ b/go-sdk/kestra_api_client/dashboards_api.go
@@ -36,12 +36,14 @@ func (a *DashboardsAPI) PreviewChart(ctx context.Context, tenant string, request
 	return doJSON[*PagedResultsMapStringObject](&a.baseAPI, ctx, "POST", tenantPath(tenant, "dashboards", "charts", "preview"), request, nil)
 }
 
-func (a *DashboardsAPI) ExportChartToCsv(ctx context.Context, tenant string, request interface{}) ([]byte, error) {
-	return a.doDownloadBytes(ctx, "POST", tenantPath(tenant, "dashboards", "charts", "export", "to-csv"), request, nil)
+func (a *DashboardsAPI) ExportChart(ctx context.Context, tenant string, request interface{}, format *string) ([]byte, error) {
+	params := buildQueryParams("format", format)
+	return a.doDownloadBytes(ctx, "POST", tenantPath(tenant, "dashboards", "charts", "export"), request, params)
 }
 
-func (a *DashboardsAPI) ExportDashboardChartDataToCSV(ctx context.Context, id, chartId, tenant string, filters interface{}) ([]byte, error) {
-	return a.doDownloadBytes(ctx, "POST", tenantPath(tenant, "dashboards", id, "charts", chartId, "export", "to-csv"), filters, nil)
+func (a *DashboardsAPI) ExportDashboardChart(ctx context.Context, id, chartId, tenant string, filters interface{}, format *string) ([]byte, error) {
+	params := buildQueryParams("format", format)
+	return a.doDownloadBytes(ctx, "POST", tenantPath(tenant, "dashboards", id, "charts", chartId, "export"), filters, params)
 }
 
 func (a *DashboardsAPI) DefaultDashboards(ctx context.Context, tenant string) (*DashboardSettings, error) {
--- a/go-sdk/test/api_dashboards_test.go
+++ b/go-sdk/test/api_dashboards_test.go
@@ -3,12 +3,37 @@ package test
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 	"testing"
 
 	"github.com/kestra-io/client-sdk/go-sdk/kestra_api_client"
 	"github.com/stretchr/testify/require"
 )
+
+// expectedExportColumns are the executionsTable*Yaml column keys, sorted. The
+// CSV header uses these raw keys (not their displayName) and the SQL layer
+// does not guarantee they come back in declaration order, so tests compare
+// against a sorted copy of the header rather than a fixed string.
+var expectedExportColumns = []string{"flow", "id", "namespace", "state"}
+
+// ionBinaryVersionMarker is the fixed 4-byte header every Amazon Ion 1.0
+// binary stream starts with; FileSerde writes ION exports in binary form.
+var ionBinaryVersionMarker = []byte{0xE0, 0x01, 0x00, 0xEA}
+
+func assertCsvHeader(t *testing.T, csv string) {
+	t.Helper()
+	header := strings.TrimRight(strings.SplitN(csv, "\n", 2)[0], "\r")
+	columns := strings.Split(header, ",")
+	sort.Strings(columns)
+	require.Equal(t, expectedExportColumns, columns)
+}
+
+func assertIonBinaryMarker(t *testing.T, result []byte) {
+	t.Helper()
+	require.GreaterOrEqual(t, len(result), 4)
+	require.Equal(t, ionBinaryVersionMarker, result[:4])
+}
 
 func dashboardYaml(title string) string {
 	return dashboardYamlWithId(randomId(), title)
@@ -24,6 +49,74 @@ timeWindow:
   max: P365D
 charts: []
 `, id, title)
+}
+
+// executionsTableDashboardYaml builds a dashboard with a single OSS-native Table
+// chart over the Executions data type, scoped to namespace so its export content
+// is deterministic even on a shared test instance with unrelated executions.
+func executionsTableDashboardYaml(dashboardId, title, chartId, namespace string) string {
+	return fmt.Sprintf(`
+id: %s
+title: %s
+description: Test dashboard
+timeWindow:
+  default: P30D
+  max: P365D
+charts:
+  - id: %s
+    type: io.kestra.plugin.core.dashboard.chart.Table
+    chartOptions:
+      displayName: Executions
+    data:
+      type: io.kestra.plugin.core.dashboard.data.Executions
+      where:
+        - field: NAMESPACE
+          type: EQUAL_TO
+          value: %s
+      columns:
+        id:
+          field: ID
+          displayName: Execution ID
+        namespace:
+          field: NAMESPACE
+          displayName: Namespace
+        flow:
+          field: FLOW_ID
+          displayName: Flow
+        state:
+          field: STATE
+          displayName: State
+`, dashboardId, title, chartId, namespace)
+}
+
+// executionsTableChartYaml is executionsTableDashboardYaml's chart block on its
+// own, usable directly in an ad-hoc PreviewRequest (no dashboard wrapper).
+func executionsTableChartYaml(chartId, namespace string) string {
+	return fmt.Sprintf(`
+id: %s
+type: io.kestra.plugin.core.dashboard.chart.Table
+chartOptions:
+  displayName: Executions
+data:
+  type: io.kestra.plugin.core.dashboard.data.Executions
+  where:
+    - field: NAMESPACE
+      type: EQUAL_TO
+      value: %s
+  columns:
+    id:
+      field: ID
+      displayName: Execution ID
+    namespace:
+      field: NAMESPACE
+      displayName: Namespace
+    flow:
+      field: FLOW_ID
+      displayName: Flow
+    state:
+      field: STATE
+      displayName: State
+`, chartId, namespace)
 }
 
 func TestDashboardsAPI_All(t *testing.T) {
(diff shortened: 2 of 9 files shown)


brian-mulier-p:
Tests will fail till [my PR](https://github.com/kestra-io/kestra/pull/17260) is merged on core as it relies on new APIs

jymaire:
> Tests will fail till [my PR](https://github.com/kestra-io/kestra/pull/17260) is merged on core as it relies on new APIs

Ok, let's wait for it before to merge this PR. Also I guess that you wait for this merge to generate JS SDK ?

jymaire:
Pushed 6f2784ba to unblock CI here — the `Test Javascript SDK` check was failing on `tsc --noEmit` (TS2304) because the three export tests still call `kestraClient.Dashboards.*`, which main dropped in #353.

While in those lines I also nested the chart under `body` (`{ body: { chart }, format }` rather than `{ chart, format }`) — the generated wrappers take the request body under `body`, as `validateDashboard` does just above, so `chart` at the top level was being dropped and the server saw an empty body. The `as any` cast hid both problems, and the `expect([400, 422]).toContain(...)` fallback meant the tests passed anyway. Revert that part if you disagree.

Context for the rerun: the upstream rename this PR tracks landed on `develop` in kestra-io/kestra@04d0a6d70 on 07-27, so the route now exists server-side. `Test Java SDK` and `Test GO SDK` both flipped to green on rerun; Python is still running. That also means the daily freshness check ([run 30333124607](https://github.com/kestra-io/client-sdk/actions/runs/30333124607)) is red on Java and Python until this lands.

Two follow-ups it interacts with:
- #372 refreshes `kestra-ee.yml` wholesale and so already carries the two `/export` paths this PR hand-edits — whichever merges second will need the other's hunk dropped.
- #373 (draft) tightens the Go `require.Contains(t, err.Error(), "")` no-ops that let `Test GO SDK` report a pass against the removed route for hours. It stays red until this PR merges, then needs a rebase over the `ExportChart` rename.

jymaire:
All checks settled. **Java, Go and Python are green**; the only red left is `Test Javascript SDK`, and it is not this PR's content:

```
✓ test_javascript_sdk/DashboardsApi.spec.ts (12 tests) 374ms
❯ test_javascript_sdk/PoliciesApi.spec.ts (16 tests | 15 failed)
```

Every dashboards test passes, including both `exportChart` CSV/ION cases and the three real-fixture ones. On Python:

```
test_export_chart_basic PASSED          test_export_dashboard_chart_csv PASSED
test_export_chart_ion PASSED            test_export_dashboard_chart_ion PASSED
```

The 15 `PoliciesApi` failures are the separate drift fixed by #372 (kestra-ee@fe7140415 moved policy create/update to `consumes = application/x-yaml`; this branch predates it). Merge #372, then bring `main` into this branch and the JS leg should go green too.

Two follow-up commits were needed beyond the `kestraClient` import fix, because each uncovered the next masked problem:

- `6f2784ba` — per-API imports, plus nesting the chart under `body`. `{ chart, format }` dropped the body, so the request was succeeding empty and `expect(result).toBeDefined()` passed.
- `66287bc7` — once the body actually reached the server the calls started erroring, and that exposed `e?.response?.status` as permanently `undefined`. The shared hey-api runtime's error interceptor assigns `status` onto the thrown error and exposes no `response`, so the `expect([400, 422]).toContain(...)` guard had never compared a real status. Now reads `e?.status` (the `FlowsApi`/`TestSuitesApi` convention) and surfaces `e.message` when the status is unexpected.

Worth a look since it is the same class of problem as #373: three separate assertions in this area were structurally incapable of failing.

jymaire:
## Review

Approach is sound and the propagation across the four SDKs is complete and consistent — the seeded-fixture rewrite of the `exportDashboardChart` tests is a real improvement over the previous tolerate-anything assertions.

I verified the plumbing before commenting: `buildQueryParams` (`base_api.go:52-62`), `queryParams` (`BaseApi.java:62-72`) and `_build_query_params` (`base_api.py:62-71`) all skip nil/None, so omitting `format` sends no `?format=`; `doDownloadBytes`/`_download_request` already accept params; all new test helpers resolve; JS flattens query params onto the top-level options object (cf. `searchDashboards({ page, size })`), so `format: 'CSV'` is the right shape; and the spec edit matches house style (`kestra-ee.yml:18641-18648` shows the generator emits optional enum query params as exactly `type: string` + `nullable: true` + `enum: [...]`).

Below are the issues worth addressing.

---

### 1. Nothing in the PR actually tests that `format` works

Every CSV/ION test pair asserts the identical thing — that the payload contains the namespace, flow ID and execution ID. A server that ignored `format` entirely and always returned CSV would pass all six new tests. The one behaviour this PR introduces is the only thing not covered.

Add a format-discriminating assertion, e.g. CSV → first line is a comma-separated header matching the chart's `displayName`s (`Execution ID,Namespace,Flow,State`); ION → the Ion shape (`{` / `id:` for text Ion, or the `\xe0\x01\x00\xea` BVM for binary). Even asserting that the two payloads *differ* would be a meaningful improvement over the current state.

Related: `exportChart_ion` (Java/Python/Go) uses an EE-only `TimeSeriesChart` and swallows 400/422. Since the OSS `Table`/`Executions` chart already works for the dashboard route, reusing it here would let `exportChart` actually return 200 and be asserted too.

### 2. Go: `require.Contains(t, err.Error(), "")` asserts nothing

In `go-sdk/test/api_dashboards_test.go`, `strings.Contains(s, "")` is always true — so both `exportChart_basic` (kept) and the new `exportChart_ion` pass unconditionally on *any* error: 401, 500, connection refused. Java and Python correctly narrow to `(400, 422)`; Go should match. The repo already has the idiom at `api_executions_test.go:254-256`:

```go
var apiErr *kestra_api_client.ApiError
require.True(t, errors.As(err, &apiErr))
require.Contains(t, []int{400, 422}, apiErr.StatusCode)
```

Pre-existing on the `_basic` test, but the PR copies it into a new one — worth fixing both while here.

### 3. Missing `### ⚠️ Signature change` section (AGENTS.md)

`AGENTS.md` requires, for every changed public method, a line under a literal `### ⚠️ Signature change` heading in the PR description. Eight methods changed both *name and arity* across four SDKs, and plugin-kestra consumes them. The body describes the rename in prose, but the heading is what the downstream non-blocking check looks for. Concretely:

```
### ⚠️ Signature change
- Java:   exportChartToCsv(tenant, request) → exportChart(tenant, request, format)
- Java:   exportDashboardChartDataToCSV(id, chartId, tenant, filters) → exportDashboardChart(id, chartId, tenant, filters, format)
- Python: export_chart_to_csv(tenant, request) → export_chart(tenant, request, format=None)
- Python: export_dashboard_chart_data_to_csv(id, chart_id, tenant, filters) → export_dashboard_chart(id, chart_id, tenant, filters, format=None)
- Go:     ExportChartToCsv(ctx, tenant, request) → ExportChart(ctx, tenant, request, format *string)
- Go:     ExportDashboardChartDataToCSV(ctx, id, chartId, tenant, filters) → ExportDashboardChart(ctx, id, chartId, tenant, filters, format *string)
- JS:     exportChartToCsv → exportChart, exportDashboardChartDataToCSV → exportDashboardChart (regenerated)
```

### 4. JS is the only SDK without a positive `exportDashboardChart` test

Go, Java and Python each got the seeded flow + execution + real chart fixture. JS got two error-swallowing `exportChart` tests and a not-found negative — no happy path. Given JS function coverage is CI-gated, porting the seeded fixture there is worth it.

### 5. JS `as any` casts remove the only compile-time proof the spec edit landed

Three `} as any)` casts mean that if `format` came out of generation under a different name (or not at all), `check:types` stays green and the request simply omits the param — and the test still passes. Dropping the casts turns `check:types` into the verification that the spec change produced the intended parameter. If a cast is needed for `body`, scope it to `body` rather than the whole options object.

---

### Nits

- **JS comment is misleading.** The 4-line past-tense comment ("*reading `e.response.status` yielded undefined and the assertion could never describe the real failure*") appears verbatim in two brand-new tests that never had a prior assertion — it reads as a changelog for a bug that isn't in this file's history. Condense to one line, or hoist a `const statusOf = (e: any) => e?.status ?? e?.response?.status` helper and drop the prose.
- **Python decode inconsistency hints at an open question.** CSV uses `result.decode()`, ION uses `result.decode("utf-8", errors="ignore")`. If the endpoint emits *binary* Ion, then Go's `string(result)` and Java's `new String(result)` substring checks work only because Ion happens to store string values inline as UTF-8 — fragile. If it's text Ion, `errors="ignore"` is unnecessary. Worth pinning down and making all three consistent.
- **Java charset.** `new String(result)` uses the platform default; benign on Java 21 (UTF-8 default) but `new String(result, StandardCharsets.UTF_8)` is the correct explicit form.
- **`format` as a raw string.** `String` / `*string` / `str` for a closed two-value enum means `"csv"` compiles and fails at runtime. Consistent with the hand-written SDKs' existing style, so not a blocker — just flagging the trade-off.

No security or performance concerns: no auth, serialization or transport changes, and the new fixtures are scoped to random namespaces so they're safe on shared instances.

jymaire (approved):
## Approve

All five points from my earlier review are addressed in `d5514d11`, and the blockers are gone.

- **Format is now actually tested.** `assertCsvHeader` compares the sorted header against `['flow','id','namespace','state']`, and `assertIonBinaryMarker` checks the 4-byte Ion BVM (Go/Java/Python) — JS checks for the U+FFFD left by UTF-8-decoding binary Ion. A server that ignored `format` would now fail these.
- **Go no-ops fixed.** Both export tests moved to `require.NoError` plus real content assertions, and `exportChart_ion` reuses the OSS `Table`/`Executions` chart so it gets a real 200 instead of swallowing 400/422. The one remaining `require.Contains(t, err.Error(), "")` is in `previewChart_basic` (`api_dashboards_test.go:334`) — pre-existing, and #373's scope.
- **`### ⚠️ Signature change`** section is in the description with all eight methods.
- **JS has happy paths** for `exportDashboardChart` in both formats, on seeded flow/execution fixtures.
- **`as any` dropped from the export calls**, so `check:types` is now the proof that the spec change produced the `format` param.

State is clean too: all four checks green on `d5514d11`, branch is 0 commits behind `main` so the `PoliciesApi` red is genuinely resolved by #372 rather than pending a rebase, the `kestra-ee.yml` hand-edit is correctly dropped now that #372 carries both `/export` paths with the `format` enum, and kestra-io/kestra#17260 merged on 07-27 so the route exists server-side.

Two things to keep in mind after merge, neither blocking:

- It is a breaking rename with no alias, so the next `main` release drops `exportChartToCsv`/`exportDashboardChartDataToCSV`. kestra-io/plugin-kestra#157 is still a draft, so nothing green breaks today, but it needs this SDK version to land.
- #373 needs a rebase over the `ExportChart` rename; it stays red until this merges.

Unrelated cleanup worth a separate issue: `go-sdk/kestra_api_client/api/openapi.yaml` and `java/java-sdk/api/openapi.yaml` still document the old `/export/to-csv` routes (vestigial — both SDKs are hand-written now), and `go-sdk/test.test` is a committed binary on `main`.

PR #324, opened 2026-07-06
feat(js-sdk): forward CSRF token from <meta> in the browser

jymaire:
## What

Forward the Kestra CSRF token as `X-CSRF-TOKEN` from the JS SDK's request interceptor.

Kestra's `CsrfTokenFilter` rejects cookie-authenticated non-safe requests (`POST`/`PUT`/`PATCH`/`DELETE`) that lack `X-CSRF-TOKEN` with a **403**. The Kestra UI exposes the token as `<meta name="csrf-token">`. Today every browser consumer (Kestra UI, plugin UIs) re-implements this lookup — the Kestra UI does it in `main.ts`, and plugins duplicate it in their composables. Moving it into `configureClient()` means every browser consumer gets it for free.

## Behaviour

- Only on unsafe methods (`POST`/`PUT`/`PATCH`/`DELETE`).
- **No-op** outside the browser (`typeof document === "undefined"` → Node, other projects).
- **Does not overwrite** a caller-provided `X-CSRF-TOKEN` (header auth stays untouched).

## Follow-ups (separate PRs)

- Kestra UI: drop the now-redundant manual CSRF interceptor in `main.ts` (removes the documented "do not call `configureClient` after this → 403" footgun).
- Plugin UIs (e.g. plugin-gcp): drop the per-call CSRF lookup from their composables.

## Tests

Added `csrfInterceptor.spec.ts`: sets the header from `<meta>` on unsafe methods, skips safe methods, preserves a caller-provided header, and is a no-op without `document`.

### DIFF
--- a/javascript/javascript-sdk/src/index.ts
+++ b/javascript/javascript-sdk/src/index.ts
@@ -53,6 +53,21 @@ export function configureClient(clientConfig: Config<ClientOptions> = {}, axiosC
                 config.headers["Accept"] = "text/csv, text/plain, text/json, application/json"
             }
         }
+
+        // Kestra's CsrfTokenFilter rejects cookie-authenticated non-safe requests (POST/PUT/PATCH/
+        // DELETE) that lack a CSRF token with a 403. In the browser the Kestra UI exposes the token
+        // as <meta name="csrf-token">, so forward it as X-CSRF-TOKEN here — every browser consumer
+        // then gets it automatically instead of re-implementing the lookup. No-op outside the browser
+        // (Node, other projects) and when the caller already set the header or uses header auth.
+        const isUnsafe = method === "post" || method === "put" || method === "patch" || method === "delete"
+        if (isUnsafe && typeof document !== "undefined" && config.headers["X-CSRF-TOKEN"] == null) {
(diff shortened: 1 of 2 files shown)


elevatebart (changes requested):
If this was necessary, how do you think EE would work today? What is the use case/issue you are facing that triggered this refactor?

The plugins should be using the client enhanced with the csrf token. It "should" not be necessary for it to be added in every plugin. The federation singleton should take care of that. If it doesn't, you will also need to figure out auth on your end... hard

jymaire:
Closing per review — the CSRF interceptor isn't needed in the SDK.

`configureClient` already wires its axios instance into the generated client singleton (`client.setConfig({axios})`), and the host UI attaches the `X-CSRF-TOKEN` request interceptor to that same instance. Since `@kestra-io/kestra-sdk` is a module-federation singleton, browser consumers (Kestra UI, plugin UIs) inherit CSRF automatically — so forwarding the `<meta>` token inside the SDK was redundant, and reading `document`/`<meta>` would leak browser-DOM assumptions into an SDK that also runs on Node and Go.

Thanks @elevatebart for the catch.

PR #325, opened 2026-07-07
agents: replace claude.md with agents.md and add some clues about JS

elevatebart:
Added a warning that everything under `javascript/javascript-sdk/src/openapi/` is auto-generated and must not be manually edited. Instructions are provided for making changes via the generator sources (`plugin.ts`, `index.ts`).


CLAUDE.md
jymaire, line 1: on plugins this is how we do, I think @fdelbrayelle might have the right explanation for this

fdelbrayelle: Yes @elevatebart since Claude Code only reads `CLAUDE.md` let's keep it with a `@AGENTS.md` it doesn't hurt anyone, does it?

elevatebart: Sounds good

fdelbrayelle (approved):

PR #335, opened 2026-07-13
fix(python): restore flow inputs on create/resume/replay execution

### LINKED ISSUE #203: SDK Flow inputs has changed. The current Python SDK docs example is stale.
### Describe the issue

The current Python SDK docs example is stale. The docs page still shows flow_id= and inputs= in Execute a flow (https://kestra.io/docs/api-reference/kestra-sdk/python-sdk#execute-a-flow)

flow inputs are passed with `additional_form_datas`, not inputs, and the flow identifier parameter is `id`, not flow_id. This is not same with other Kestra standards and naming convention. Flow inputs should be called `inputs`

### Environment

- Kestra Version: develop

jymaire:
## Problem

The 1.3.0 rewrite that hand-wrote the Python SDK API classes (#237) dropped the ability to pass **flow inputs** to executions. Reported in #203 and Pylon #2038 (HeartFlow): after upgrading `kestrapy` from 1.0.10 → 1.3.0, `create_execution` lost its `additional_form_datas` parameter and gained no replacement, so there was no way to start a flow with inputs.

Auditing the SDK for the same class of regression (a `multipart/form-data` request body silently dropped during the rewrite) turned up **two more** endpoints with the identical bug.

## Fixes

All three `multipart/form-data` inputs endpoints now accept an `inputs` mapping, encoded via a shared `_build_inputs_multipart` helper:

- **`create_execution`** — start a flow with inputs (the reported issue). Uses `inputs=`, matching Kestra's naming convention (per #203).
- **`resume_execution`** — supply the `onResume` inputs declared by a `Pause` task. Previously sent no body, so resuming a pause with required inputs always failed.
- **`replay_execution_with_inputs`** — override flow inputs on replay. Previously sent no inputs, making it behave identically to `replay_execution` despite its name.

Value conventions for the `inputs` mapping:
- `str` → sent verbatim
- `bytes` / `(filename, content[, content_type])` tuple → file part (FILE inputs)
- anything else (`bool`, `int`, `dict`, `list`, …) → JSON-encoded (`True`→`"true"`, `[1,2]`→`"[1, 2]"`)

## Tests

Added integration tests that assert real values:
- `test_create_execution_with_inputs` — runs a flow with a `STRING` input, asserts the echoed `inputs.greeting`
- `test_resume_execution_with_inputs` — pauses on a `Pause` task with a **required** `onResume` input, resumes with it, asserts SUCCESS
- `test_replay_execution_with_inputs_overrides_inputs` — replays overriding an input, asserts the new value

## Audit note

I audited the full hand-written SDK for the same class of regression (dropped bodies, wrong content-types, dropped query params, dropped methods, wrong HTTP verbs), cross-referencing both the OpenAPI spec and the pre-#237 generated code. The **only** genuine regressions were these three dropped multipart bodies. Query-param and method-coverage differences are all Kestra 2.0-vs-1.3 version drift or renames, not rewrite regressions.

Closes #203

### DIFF
--- a/python/python-sdk/kestrapy/api/executions_api.py
+++ b/python/python-sdk/kestrapy/api/executions_api.py
@@ -1,3 +1,4 @@
+import json
 from typing import Any, Dict, Generator, List, Optional
 
 from kestrapy.base_api import BaseApi
@@ -30,6 +31,7 @@ class ExecutionsApi(BaseApi):
         tenant: str,
         namespace: str,
         id: str,
+        inputs: Optional[Dict[str, Any]] = None,
         labels: Optional[List[str]] = None,
         wait: Optional[bool] = None,
         revision: Optional[int] = None,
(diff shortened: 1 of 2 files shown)


jymaire:
### Backward compatibility with 1.3.0

`inputs` is appended at the **end** of each signature, so the change is purely additive — every existing 1.3.0 positional argument keeps its position:

- `create_execution(tenant, namespace, id, labels, wait, revision, schedule_date, breakpoints, kind, inputs=None)`
- `replay_execution_with_inputs(execution_id, tenant, task_run_id, revision, breakpoints, inputs=None)`
- `resume_execution(execution_id, tenant, inputs=None)`

Existing 1.3.0 callers (positional or keyword) are unaffected; `inputs` is opt-in.

Malaydewangan09 (approved):

PR #336, opened 2026-07-13
fix(python/1.3): restore flow inputs on create/resume/replay execution

jymaire:
## Problem

kestrapy **1.3.x** dropped the ability to pass **flow inputs** to executions. Reported in #203 and Pylon #2038 (HeartFlow): after upgrading `kestrapy` 1.0.10 → 1.3.0, `create_execution` lost its `additional_form_datas` parameter and gained no replacement — there was no way to start a flow with inputs.

The hand-written SDK rewrite reimplemented three `multipart/form-data` endpoints as bodyless requests. All three are fixed here.

> This is the **1.3 line** fix (the branch 1.3.x is released from). The equivalent fix for `main`/2.0 is #335 — the two differ only in the endpoint paths (1.3 has no `/actions/` segment).

## Fixes

An `inputs` mapping, sent as `multipart/form-data` via a shared `_build_inputs_multipart` helper:

| method | 1.3 path |
|---|---|
| `create_execution` | `POST executions/{namespace}/{id}` |
| `resume_execution` | `POST executions/{executionId}/resume` |
| `replay_execution_with_inputs` | `POST executions/{executionId}/replay-with-inputs` |

Value conventions: `str` sent verbatim; `bytes` / `(filename, content[, content_type])` tuple → file part (FILE inputs); anything else JSON-encoded (`True`→`"true"`, `[1,2]`→`"[1, 2]"`).

## Backward compatibility

`inputs` is appended at the **end** of every signature, so the change is purely additive — all existing 1.3.x positional arguments keep their position, `inputs` is opt-in. `resume_execution` keeps its `None` return.

## Tests

- `test_create_execution_with_inputs` — flow with a `STRING` input, asserts echoed `inputs.greeting`
- `test_resume_execution_with_inputs` — pauses on a `Pause` task with a **required** `onResume` input, resumes with it, asserts SUCCESS
- `test_replay_execution_with_inputs_overrides_inputs` — replays overriding an input, asserts the new value

Ref Pylon #2038. Ships to customers in the next 1.3 patch (1.3.2).

### DIFF
--- a/python/python-sdk/kestrapy/api/executions_api.py
+++ b/python/python-sdk/kestrapy/api/executions_api.py
@@ -1,3 +1,4 @@
+import json
 from typing import Any, Dict, Generator, List, Optional
 
 from kestrapy.base_api import BaseApi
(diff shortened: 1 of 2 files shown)


jymaire:
### QA on live Kestra EE `v1.3.27-no-plugins`

Verified against a real 1.3.27 instance (`kestra-ee:v1.3.27-no-plugins`, tenant `default`), reproducing the customer bug on the released SDK, then confirming the fix.

**Before (released `releases/v1.3.x`):** flow with a required `STRING` input
- `create_execution(..., inputs=...)` → `TypeError: unexpected keyword argument 'inputs'` — no way to pass inputs (customer's exact complaint)
- `create_execution(...)` without inputs → `422 Invalid entity: Missing required input:greeting`

**After (this branch):**
- `create_execution(..., inputs={"greeting":"hello"})` → execution `SUCCESS`, `inputs={'greeting': 'hello'}` ✅
- `resume_execution(id, inputs={"reason":"go"})` on a `Pause` task with a required `onResume` input → `SUCCESS` ✅
- `replay_execution_with_inputs(id, inputs={"greeting":"world"})` → new execution `SUCCESS`, `inputs={'greeting': 'world'}` (override applied) ✅

All three multipart-inputs endpoints now work end-to-end on 1.3.

Malaydewangan09 (approved):
LGTM, but CI seems to be failing

jymaire:
> LGTM, but CI seems to be failing

Seems to be some OOM, I try to restart it

PR #338, opened 2026-07-13
test(javascript): cover long-tail SDK functions (#332)

jymaire:
Part of #332 — raising JS SDK function test coverage toward the ≥90% goal.

Adds tests for **19** previously-untested generated SDK wrappers, prioritizing functions that can be driven reliably against a live Kestra instance (create → read/update/delete round-trips and list endpoints), no source changes.

| File | Functions covered |
|------|-------------------|
| `Blueprints` | `internalBlueprint`, `internalBlueprintFlow`, `updateInternalBlueprints`, `deleteInternalBlueprints` |
| `Executions` | `searchExecutionsByFlowId`, `listFlowExecutionsByNamespace`, `listExecutableDistinctNamespaces`, `findDistinctFieldValues`, `evalExpression` |
| `Namespaces` | `exportPluginDefaults`, `importPluginDefaults` |
| `Misc` | `listPermissions`, `tenantUsage` |
| `WorkerGroups` | `capacity`, `listRunningWorkers` |
| `ServiceAccount` | `listServiceAccountsForTenant`, `listApiTokensForServiceAccountWithTenant`, `createApiTokensForServiceAccountWithTenant`, `deleteApiTokenForServiceAccountWithTenant` |

### Notes
- +214 lines, tests only (well under the 500-line target).
- `npm run check:types` passes.
- The coverage ratchet floor in `vitest.config.ts` is left at `functions: 69`. It only guards against regressions, so it stays safe as coverage rises; bump it to the new measured floor once the CI coverage comment reports the real number.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #332, opened 2026-07-09
Increase JavaScript SDK test coverage (function tests: 71.1% → ≥90%)


jymaire:
## Summary

Sub-issue of #331. Raise the **JavaScript SDK function test coverage** to the agreed target (**≥ 90%**) and turn the existing CI coverage report into a real regression **gate**.

Unlike the hand-written Go/Python/Java SDKs, the JS SDK is **generated** and already wraps the full API surface — so the gap here is not *surface* coverage but *function test* coverage: **71.1% (308 / 433 functions)**, leaving **125 functions untested**.

## Current state

- CI already measures and reports coverage: `.github/scripts/coverage-comment.mjs` (wired into `.github/workflows/javascript-sdk.yml`) reads the istanbul `coverage-final.json` and posts a sticky PR comment listing every uncovered function in `openapi/sdk/`.
- **But it does not gate.** The vitest coverage `thresholds` block is commented out in `javascript/test_javascript_sdk/vitest.config.ts`:
  ```ts
  // TODO: reenable this once we have the necessary coverage
  // thresholds: {
  //     perFile: true,
  //     functions: 60,
  // }
  ```
  So today the number can silently regress — the PR comment is informational only.
- Some APIs are intentionally excluded from the denominator (`IGNORE_APIS` in `vitest.config.ts`: `Ai`, `Login`, `TenantAccess`, `WorkerAuth`, plus their `*Admin` variants).

## Scope

- [ ] Write tests covering the ~125 untested functions in `javascript-sdk/src/openapi/sdk/`, prioritising the largest resource groups (executions, namespaces, instance, flows, apps, triggers…) to push coverage from **71.1% → ≥ 90%**.
- [ ] Once coverage is high enough, **re-enable the vitest `thresholds` gate** (`perFile` + `functions`) so PRs that drop coverage fail CI, not just get a comment. Set the threshold to the agreed target.
- [ ] Confirm the sticky PR comment and the gate agree on the same numerator/denominator (same `IGNORE_APIS` exclusions).

## How to reproduce locally

```bash
./generate-sdks.sh javascript        # generate + build the SDK first
cd javascript
npm run test --workspace test_javascript_sdk -- --coverage
```

The PR comment lists exactly which functions are uncovered, per file — use it as the worklist.

## Acceptance Criteria

- [ ] JavaScript function test coverage ≥ 90% (currently 71.1%)
- [ ] vitest coverage `thresholds` re-enabled so regressions fail CI
[...]

PR #339, opened 2026-07-15
test(javascript): cover read-only Plugins SDK functions (#332)

jymaire:
## What

Adds integration tests for previously-untested read-only functions in the Plugins SDK domain, continuing the coverage work for #332.

New tests (`PluginsApi.spec.ts`):
- `propertiesFromType` — asserts the property map contains base Task `id`/`type`
- `schemasFromType` — asserts the flow JSON schema carries a `properties` map
- `pluginIcon` — asserts a core plugin ships a non-null icon
- `pluginVersions` — asserts the endpoint echoes the requested class in `type`
- `pluginDocumentationFromVersion` — sources a real version from `pluginVersions`, asserts rendered markdown documents the Log task
- `pluginUiManifest` — asserts the manifest map is present but empty for core Log

Plus one deferred `it.skip('pluginUi')` — no plugin ships a UI bundle in the test environment, so there is no valid `{group, path}` to fetch.

## Notes

- Assertions check real response values (no `toBeDefined()`/`toBeGreaterThan(0)` placeholders), following the conventions from #333 / #338.
- The ratchet gate (`vitest.config.ts` `thresholds.functions`) is untouched; bump it once CI reports the new coverage floor.

### DIFF
--- a/javascript/test_javascript_sdk/PluginsApi.spec.ts
+++ b/javascript/test_javascript_sdk/PluginsApi.spec.ts
@@ -59,20 +59,19 @@ describe('PluginsApi', () => {
     });
 
     it('propertiesFromType: returns the JSON-schema properties for a type', async () => {
-        // The response is an untyped JSON Schema (Draft-7) document whose
-        // top-level layout is not stable across types/versions (it mixes string
-        // fields like `$schema`/`$ref` with nested objects), so assert it is a
-        // populated object rather than a specific key.
         const result = await kestraClient.Plugins.propertiesFromType({ type: 'TASK' });
-        expect(Array.isArray(result)).toBe(false);
-        expect(result).not.toEqual({});
+        // A JSON Schema document: a `$schema` dialect marker plus the
+        // `properties`/`required` describing every task property.
+        expect(result).toMatchObject({ $schema: expect.stringContaining('json-schema.org') });
+        expect(result).toHaveProperty('properties');
     });
 
     it('schemasFromType: returns the JSON schema for a type', async () => {
-        // Same untyped Draft-7 schema document as propertiesFromType.
         const result = await kestraClient.Plugins.schemasFromType({ type: 'FLOW' });
-        expect(Array.isArray(result)).toBe(false);
-        expect(result).not.toEqual({});
+        // A JSON Schema document: a `$schema` dialect marker, a top-level `$ref`,
+        // and the `definitions` it resolves into.
+        expect(result).toMatchObject({ $schema: expect.stringContaining('json-schema.org') });
+        expect(result).toHaveProperty('$ref');
     });
 
     it('pluginIcon: returns the icon for a plugin class', async () => {
--- a/javascript/test_javascript_sdk/PluginsApi.spec.ts
+++ b/javascript/test_javascript_sdk/PluginsApi.spec.ts
@@ -108,12 +108,4 @@ describe('PluginsApi', () => {
         // ships no custom UI, so the manifest map is present but empty.
         expect(result.manifest).toEqual({});
     });
-
-    // Serves a plugin's bundled UI static asset. No plugin ships a UI bundle in
-    // the default test environment, so there is no valid {group, path} to fetch.
-    it.skip('pluginUi: serves a plugin UI static asset', async () => {
-        const result = await kestraClient.Plugins.pluginUi({ group: 'io.kestra.plugin.core', path: 'index.js' });
-        // The endpoint streams the asset back as a Blob (File extends Blob).
-        expect(result).toBeInstanceOf(Blob);
-    });
 });
--- a/javascript/test_javascript_sdk/PluginsApi.spec.ts
+++ b/javascript/test_javascript_sdk/PluginsApi.spec.ts
@@ -59,7 +59,7 @@ describe('PluginsApi', () => {
     });
 
     it('propertiesFromType: returns the JSON-schema properties for a type', async () => {
-        const result = await kestraClient.Plugins.propertiesFromType({ type: 'TASK' });
+        const result = await Plugins.propertiesFromType({ type: 'TASK' });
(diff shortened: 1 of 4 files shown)


javascript/test_javascript_sdk/QuotasApi.spec.ts
```diff
@@ -0,0 +1,12 @@
...
+import './CommonTestSetup.js';
+import * as Quotas from '@kestra-io/kestra-sdk/quotas';
+
+describe('QuotasApi', () => {
+    it('search: lists quota limits for the tenant', async () => {
+        // Read-only, tenant-scoped endpoint with no create API, so a tenant with
+        // no configured limits returns an empty array. Assert the shape.
+        const result = await Quotas.search();
+        expect(Array.isArray(result)).toBe(true);
+    });
```
elevatebart, line 14: Could we set some quotas and then read their values?

elevatebart: Maybe in the config in `test-utils/docker-setup/application.yml`

elevatebart (approved):
One NIT on quotas (same goes for reusable-inputs)

javascript/test_javascript_sdk/QuotasApi.spec.ts
jymaire: Good call. Two different mechanisms here:

**Quotas** — `search` (`GET /quota-limits`) returns `QuotaLimit` *usage counters* (`{tenantId, namespace, flowId, id, start, count}`), materialized lazily as quota-consuming activity happens — not the configured limits, so a fresh tenant genuinely returns `[]`. The configurable side is `Tenant.quotas` (`Quota{duration, limit, behavior}`), not `application.yml`. Added a test that creates a tenant carrying a quota and asserts it survives the GET round-trip, so we now assert a real value. Kept the shape check on `search` since there's no usage to count in the test env.

**Reusable inputs** — no tenant/config analog (blocks live in namespace YAML), and the only write path (`createOrUpdate`) is the JSON-vs-YAML content-type bug tracked in #340, so seeding a block isn't possible yet. Left as the documented shape assertion until #340 lands.

elevatebart: Update/rebase the main branch into this one and check again.

jymaire: Correction / update after actually spinning up a `kestra-ee:develop` instance (the CI target for this PR) and probing it:

I tried both the round-trip and the "run a flow then read the usage" idea, and **neither works through the API**:

- **Quotas don't round-trip.** `PUT /tenants/{id}` (and create) with a `quotas` array returns a Tenant with **no** `quotas` field — it's never echoed back. So a round-trip assertion always fails.
- **Tenant quotas have no observable effect via the API.** With `{duration: PT1H, limit: 1, behavior: FAIL}` set on the tenant, I ran the same flow 3× — all succeeded (no gating), and `GET /quota-limits` stayed `[]` throughout.

So there's no real value to assert beyond the shape: `search()` returns `[]` with 200 even after activity, and the only config surface (`Tenant.quotas`) doesn't feed it. I've reverted the round-trip test and kept the `Array.isArray` check, now with a comment documenting the verified behavior. Same conclusion stands for reusable-inputs (blocked by #340).

If we want genuine `quota-limits` coverage later, it looks like it'd need a backend mechanism that isn't reachable via the current SDK surface — worth a separate issue rather than blocking this one.

jymaire: Final correction — I was wrong in my previous message, apologies for the noise. My "quotas don't round-trip" probe was run against a **stale** local `kestra-ee:develop` image. After pulling a fresh `develop`, creating a tenant with a `Quota` **does** round-trip:

```
POST /api/v1/tenants {"quotas":[{"duration":"PT1H","limit":1000,"behavior":"FAIL"}]}
GET  /api/v1/tenants/qrt -> ...,"quotas":[{"duration":"PT1H","limit":1000,"behavior":"FAIL"}]
```

CI confirms it too (run on 67d49f23: `QuotasApi.spec.ts (2 tests) ✓`). So the round-trip test is valid and asserts a real quota value — exactly your NIT. I've restored it.

(`search()`/`quota-limits` still only returns usage counters, so that one keeps the shape assertion; and reusable-inputs stays shape-only, blocked by #340.)

javascript/test_javascript_sdk/PluginsApi.spec.ts
```diff
@@ -57,4 +57,55 @@ describe('PluginsApi', () => {
         const result = await Plugins.pluginDocumentation({ cls: 'io.kestra.plugin.core.log.Log' });
         expect(result).toBeDefined();
     });
+
+    it('propertiesFromType: returns the JSON-schema properties for a type', async () => {
+        const result = await kestraClient.Plugins.propertiesFromType({ type: 'TASK' });
```
elevatebart: The kestraClient has disapeared. you should call Plugins directly

```suggestion
        const result = await Plugins.propertiesFromType({ type: 'TASK' });
```

javascript/test_javascript_sdk/ReusableInputsApi.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+        // least one reusable inputs block. A tenant with none returns an empty
+        // array, so assert the response shape.
+        expect(Array.isArray(result)).toBe(true);
+    });
+
+    // `createOrUpdate` (and `revisions`, which needs a block to exist) are not
+    // covered: the endpoint consumes application/x-yaml, but the generated SDK
+    // wrapper hardcodes Content-Type: application/json, so the YAML body is
+    // rejected (HTTP 500). Tracked in #340 — re-enable once the generator emits
+    // the correct content type for YAML request bodies.
```
elevatebart: Can we reevaluate the testing of `createOrUpdate` now that 340 is solved?

PR #342, opened 2026-07-16
fix(javascript): share client interceptors and axios-parity gaps with useClient()

elevatebart:

`useClient()`'s ad-hoc HTTP methods bypassed `configureClient()`'s interceptor pipeline entirely — a `client.interceptors.*.use()` registered by an app (CSRF header, 401 handling) silently never applied to `useClient()` calls, even though pre-#225 both paths shared one axios instance. `useClient()` also lost `params`, `responseType: "blob"`, `timeout`, and `request.responseURL` versus axios.

`axiosLikeRequest` now builds a real `Request`, runs it through the same `client.interceptors.request/response/error` arrays `configureClient()` populates, then fetches. Errors get the same status/message normalization as generated calls, plus a `.response` for backward compat with `err.response.data` call sites.

- `tsc --noEmit` + `npm run build`: clean, including `test_javascript_sdk`'s typecheck against the rebuilt package.
- Manual runtime check against the built `dist/index.js` with a mocked `fetch`: confirms interceptor sharing, `params` serialization, error normalization, and FormData content-type handling all work. (Full `npm test` needs a live Kestra instance, not available here.)

### Setup Instructions

None.

### DIFF
--- a/javascript/javascript-sdk/src/index.ts
+++ b/javascript/javascript-sdk/src/index.ts
@@ -30,9 +30,7 @@ function serializeQueryValue(val: unknown) {
 interface AxiosLikeConfig {
     params?: Record<string, unknown>
     headers?: Record<string, string>
-    /** Defaults to auto-detecting from the response Content-Type (json, else text). */
     responseType?: "json" | "text" | "blob"
-    /** Aborts the request after this many ms. */
     timeout?: number
     validateStatus?: (status: number) => boolean
     [key: string]: any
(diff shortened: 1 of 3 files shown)


javascript/javascript-sdk/src/index.ts
elevatebart, line 1: Make sure comments are kept to a minimum

elevatebart: Trimmed in 69a5ce8 — cut the multi-line JSDoc/comments down to one-liners, kept only the non-obvious "why" ones.

---

jymaire (approved):

PR #345, opened 2026-07-17
fix: bring back inputs for resumeExecution method

Skraye:
Commit can then be backported directly to 1.3 branch, there should be no conflict

### DIFF
--- a/java/java-sdk/java-sdk-test/src/test/java/io/kestra/sdk/api/ExecutionsApiTest.java
+++ b/java/java-sdk/java-sdk-test/src/test/java/io/kestra/sdk/api/ExecutionsApiTest.java
@@ -59,6 +59,20 @@ public class ExecutionsApiTest {
                 """.formatted(id, ns);
     }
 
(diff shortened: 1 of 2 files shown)


java/java-sdk/java-sdk-test/src/test/java/io/kestra/sdk/api/ExecutionsApiTest.java
```diff
@@ -606,6 +620,29 @@ void pauseExecution_notPaused() throws ApiException {
         assertThat(resumed.getId()).isEqualTo(executionId);
     }
 
+    @Test
+    @Disabled("Kestra 2.0: returns 404 'Requested Flow is not found' — execution action now requires the source flow to exist")
```
jymaire: why is it added and disabled ? `main` is SDK should be compatible with 2.0

jymaire:
```suggestion
```

fdelbrayelle (approved):

PR #346, opened 2026-07-17
docs(agents): notify on suspicious signature/BC changes without blocking

fdelbrayelle:
## Summary

- AGENTS.md now instructs contributors to diff and note public-method signature changes on hand-written/generated API classes (`### ⚠️ Signature change`, old → new) in the PR description — non-blocking, doesn't gate deliberate changes.

Context: `resumeExecution`'s `inputs` param was silently dropped during a hand-written Java rewrite (unit tests stayed green since nothing exercised it), then propagated to plugin-kestra on its next SDK bump. Companion check added to plugin repos in kestra-io/engineering-ai-hub#85. See https://kestra-io.slack.com/archives/C08LERML509/p1784281489868259.

## Test plan

- [ ] Docs-only change, no build/test impact

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -52,7 +52,7 @@ Tests require a running Kestra instance at `http://localhost:9903` with credenti
 
 ## Signature Changes — Notify, Don't Block
 
-Java, and now parts of Python (`executions_api.py`), are **hand-written**, not generated — a manual rewrite can silently drop a parameter with no generator to catch it. This already happened to `resumeExecution`'s `inputs` param in Java, and independently to `create_execution` / `resume_execution` / `replay_execution_with_inputs` in Python.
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -49,3 +49,15 @@ Tests require a running Kestra instance at `http://localhost:9903` with credenti
 ./generate-sdks.sh python       # Python only
 ./generate-sdks.sh go           # Go only
 ```
+
+## Signature Changes — Notify, Don't Block
+
+Java, and now parts of Python (`executions_api.py`), are **hand-written**, not generated — a manual rewrite can silently drop a parameter with no generator to catch it. This already happened to `resumeExecution`'s `inputs` param in Java, and independently to `create_execution` / `resume_execution` / `replay_execution_with_inputs` in Python.
```
jymaire:
```suggestion
Java, Python and Go are **hand-written**, not generated — a manual rewrite can silently drop a parameter with no generator to catch it. This already happened to `resumeExecution`'s `inputs` param in Java, and independently to `create_execution` / `resume_execution` / `replay_execution_with_inputs` in Python.
```

PR #354, opened 2026-07-20
fix(javascript): send application/x-yaml Content-Type for YAML-source request bodies

### LINKED ISSUE #340: JS SDK: YAML-body endpoints send Content-Type: application/json (rejected with HTTP 500)
## Summary

Several generated JavaScript SDK wrappers target endpoints that consume `application/x-yaml`, but the generated operation hardcodes `Content-Type: application/json`. The server then rejects the YAML body, returning **HTTP 500**.

## Reproduction

`ReusableInputs.createOrUpdate` (`PUT /api/v1/{tenant}/namespaces/{namespace}/reusable-inputs/{id}`):

```ts
await ReusableInputs.createOrUpdate({
  namespace, id,
  body: 'id: my-block\nnamespace: my-ns\ninputs:\n  - id: in\n    type: STRING',
});
// -> AxiosError: Request failed with status code 500
```

## Root cause

The OpenAPI spec declares the request body as:

```yaml
requestBody:
  content:
    application/x-yaml:
      schema: { type: string }
    text/plain:
      schema: { type: string }
```

But the generated wrapper (`javascript-sdk/src/openapi/sdk.gen.ts`) emits:

```ts
headers: {
    'Content-Type': 'application/json',   // <-- should be application/x-yaml for this body
    ...options?.headers,
    ...params.headers
}
```

So a YAML string is sent with a JSON content type and the controller fails to deserialize it.

## Scope

The spec contains ~20 `application/x-yaml` request bodies, so this likely affects multiple wrappers (any YAML-source create/update endpoint), not just reusable inputs. The generator should pick the declared YAML (or text/plain) content type for these bodies rather than defaulting to `application/json`.

## Fix location

This is generated code and must be fixed in the generator, not by hand:
- `javascript/javascript-sdk/heyapi-sdk-plugin/plugin.ts`

## Workaround

Callers can override the header per request:

```ts
await ReusableInputs.createOrUpdate({ namespace, id, body }, {
  headers: { 'Content-Type': 'application/x-yaml' },
});
```

## Context

Found while adding SDK function-coverage tests (#332). The `ReusableInputs` `createOrUpdate`/`revisions` cases are currently omitted from the test suite because of this bug (see `test_javascript_sdk/ReusableInputsApi.spec.ts`).

elevatebart:
Closes #340.

- **Root cause:** hey-api resolves a request body to the `application/json` variant whenever one is declared, so `ReusableInputs.createOrUpdate` — whose spec declares `application/x-yaml`, `text/plain` and a string-schema `application/json` (mirroring the controller's `@Consumes`) — was generated with a hardcoded `Content-Type: application/json`, and the server rejected the raw YAML body with HTTP 500.
- **Fix:** a `parser.patch.operations` callback (`heyapi-sdk-plugin/patch.ts`, wired in `openapi-ts.config.ts`) drops the JSON variant and puts `application/x-yaml` first for request bodies whose YAML and JSON variants are both plain strings — so `sdk.gen.ts` and the wrappers emit `Content-Type: application/x-yaml`; structured-schema bodies like `validateTask` (all `type: object`) are untouched, and the shared `kestra-ee.sanitized.yml` used by the Go generator is not modified.
- **QA:** regenerated the SDK (exactly one operation affected), `npm run check:types` passes, and a mocked-`fetch` run confirms `createOrUpdate` now sends raw YAML as `application/x-yaml`, `createFlow`/`validateTask` are unchanged, and caller-supplied `Content-Type` overrides still win.

https://claude.ai/code/session_016bZdkgLe3pk2cpmUguLG2Y

### DIFF
--- a/javascript/javascript-sdk/heyapi-sdk-plugin/patch.ts
+++ b/javascript/javascript-sdk/heyapi-sdk-plugin/patch.ts
@@ -1,22 +1,6 @@
 /**
- * `parser.patch.operations` callback that forces the generator to emit
- * `Content-Type: application/x-yaml` for YAML-source request bodies.
- *
- * Some endpoints (e.g. `PUT .../reusable-inputs/{id}`) accept a raw YAML document and declare
- * `application/x-yaml`, `text/plain` AND `application/json` request-body variants, all with a
- * plain `type: string` schema. hey-api resolves a request body to the JSON variant whenever one
- * is declared (`contents.find(json) || contents[0]` in its requestBody parser), so the generated
- * operation hardcodes `Content-Type: application/json` and the server fails to deserialize the
- * raw YAML string (HTTP 500 — kestra-io/client-sdk#340).
- *
- * For these string-schema YAML bodies, drop the `application/json` variant and reorder the
- * remaining content so `application/x-yaml` comes first — the parser then resolves the body to
- * the YAML variant regardless of spec ordering, and both `sdk.gen.ts` operations and the
- * generated wrappers send the correct content type.
- *
- * Bodies whose JSON variant carries a structured (non-string) schema (e.g. `validateTask`,
- * whose variants are all `type: object`) are left untouched: JSON is a legitimate
- * representation there and the SDK serializes those bodies as JSON objects.
+ * hey-api prefers a declared `application/json` request-body variant, mislabeling raw YAML source bodies (#340).
+ * For string-schema YAML bodies, drop the JSON variant and put `application/x-yaml` first so the parser picks it.
  */
 const YAML_MEDIA_TYPE = "application/x-yaml";
 const JSON_MEDIA_TYPE = "application/json";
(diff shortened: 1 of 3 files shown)


javascript/javascript-sdk/heyapi-sdk-plugin/patch.ts
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * For these string-schema YAML bodies, drop the `application/json` variant and reorder the
+ * remaining content so `application/x-yaml` comes first — the parser then resolves the body to
+ * the YAML variant regardless of spec ordering, and both `sdk.gen.ts` operations and the
+ * generated wrappers send the correct content type.
+ *
+ * Bodies whose JSON variant carries a structured (non-string) schema (e.g. `validateTask`,
+ * whose variants are all `type: object`) are left untouched: JSON is a legitimate
+ * representation there and the SDK serializes those bodies as JSON objects.
+ */
```
elevatebart, line 4: This comment should be less than 3 lines long

jymaire (approved):

PR #368, opened 2026-07-24
feat(js): detect EE-only routes via a generated registry, throw EnterpriseFeatureError on 404

elevatebart:

Companion PR: kestra-io/kestra#17622 (edition + route-matched 404 headers, `SdkVersionMismatchError`), merged as `2d151f6` and released as `@kestra-io/hey-api-plugin@0.4.0`.

Replaces the abandoned `client-sdk#365`.

- New `enterprise-only-routes-plugin.ts`: local hey-api codegen plugin reading the `x-kestra: {edition: ee}` tag off each operation, baking a route registry into the generated SDK.
- `src/index.ts`: builds the `enterpriseFeature` config (with real per-feature docs slugs) and passes it to `createConfigureClient`, so a 404 on a known EE-only route throws `EnterpriseFeatureError`. Also re-exports `SdkVersionMismatchError`.
- Pins `@kestra-io/hey-api-plugin` to `0.4.0`, carrying the `X-Kestra-Route-Matched`/`X-Kestra-Edition` header disambiguation and `SdkVersionMismatchError`. The pin was written as `0.3.0` while that PR was open; `develop` shipped its own `0.3.0` in the meantime (kestra-io/kestra#19632), so the version moved up.
- `EnterpriseFeatureError`'s `docsUrl` and `contactSalesUrl` are UTM-tagged (`utm_source=sdk&utm_medium=referral&utm_campaign=ee-feature-error`, feature key in `utm_content`), following the same convention the app uses on its own kestra.io links.

Tests added: `useClientEnterpriseFeature.spec.ts`, `featureToSlugMap.spec.ts`. Verified after the `0.4.0` release: `npm install` resolves the pinned tarball, the full generation pipeline (customizer → `openapi-ts` → `tsdown` build) runs end to end against the real spec, and `check:types` is clean.

### DIFF
--- a/javascript/javascript-sdk/src/index.ts
+++ b/javascript/javascript-sdk/src/index.ts
@@ -1,9 +1,12 @@
 import { client } from "./openapi/client.gen"
 import { formDataBodySerializer } from "./openapi/client"
 import type { ResolvedRequestOptions } from "./openapi/client"
-import { createConfigureClient } from "@kestra-io/hey-api-plugin/runtime"
+import { createConfigureClient, EnterpriseFeatureError } from "@kestra-io/hey-api-plugin/runtime"
+import type { EnterpriseFeatureConfig } from "@kestra-io/hey-api-plugin/runtime"
+import { ENTERPRISE_ONLY_ROUTES_JSON } from "./openapi/sdk/enterpriseOnlyRoutes.gen"
 
 export * from "./openapi/index"
+export { EnterpriseFeatureError }
 
 declare global {
     interface Window {
@@ -141,7 +144,28 @@ const axiosLikeClient = {
     patch: <T = any>(url: string, data?: any, config?: AxiosLikeConfig) => axiosLikeRequest<T>("PATCH", url, data, config),
 } as const
 
-export const configureClient = createConfigureClient(client, formDataBodySerializer)
+// Route registry baked in at generation time from the EE spec's `x-kestra: {edition: ee}` tags
+// (see enterprise-only-routes-plugin.ts) — lets the shared error interceptor tell a 404 on an
+// EE-only route (e.g. listAuditLogs on an OSS server) apart from a genuinely-missing resource.
+const enterpriseOnlyRoutes: Record<string, { feature: string }> = JSON.parse(ENTERPRISE_ONLY_ROUTES_JSON)
+
+// docsUrl/contactSalesUrl are intentionally generic (not a per-feature marketing URL table) —
+// pending the real per-feature marketing links, a generic destination is more honest than a
+// guessed one.
+//
+// NOTE: matchRoute + 404 alone can misfire on a real EE server (a genuine not-found on an
+// EE-only route with a path param, or an SDK/server version mismatch) — see kestra's
+// ErrorController/EditionHeaderFilter (X-Kestra-Entity / X-Kestra-Edition response headers) and
+// @kestra-io/hey-api-plugin's runtime.ts, which reads them to disambiguate. That header-reading
+// logic ships in @kestra-io/hey-api-plugin >= 0.3.0 (not yet released as of this pin at 0.2.0) —
+// re-pin the devDependency above once it is, no other change needed here.
+const enterpriseFeature: EnterpriseFeatureConfig = {
+    matchRoute: (method, path) => enterpriseOnlyRoutes[`${method} ${path}`],
+    docsUrl: (feature) => `https://kestra.io/docs/enterprise-edition/${feature}`,
+    contactSalesUrl: () => "https://kestra.io/demo",
+}
+
+export const configureClient = createConfigureClient(client, formDataBodySerializer, enterpriseFeature)
 
 /**
  * Set a mock client instance controlled in tests
--- a/javascript/javascript-sdk/src/index.ts
+++ b/javascript/javascript-sdk/src/index.ts
@@ -144,21 +144,11 @@ const axiosLikeClient = {
     patch: <T = any>(url: string, data?: any, config?: AxiosLikeConfig) => axiosLikeRequest<T>("PATCH", url, data, config),
 } as const
 
-// Route registry baked in at generation time from the EE spec's `x-kestra: {edition: ee}` tags
-// (see enterprise-only-routes-plugin.ts) — lets the shared error interceptor tell a 404 on an
-// EE-only route (e.g. listAuditLogs on an OSS server) apart from a genuinely-missing resource.
+// Route registry baked in at generation time from the EE spec's x-kestra tags; see
+// enterprise-only-routes-plugin.ts. docsUrl/contactSalesUrl are generic pending real links.
 const enterpriseOnlyRoutes: Record<string, { feature: string }> = JSON.parse(ENTERPRISE_ONLY_ROUTES_JSON)
 
-// docsUrl/contactSalesUrl are intentionally generic (not a per-feature marketing URL table) —
-// pending the real per-feature marketing links, a generic destination is more honest than a
-// guessed one.
-//
-// NOTE: matchRoute + 404 alone can misfire on a real EE server (a genuine not-found on an
-// EE-only route with a path param, or an SDK/server version mismatch) — see kestra's
-// ErrorController/EditionHeaderFilter (X-Kestra-Entity / X-Kestra-Edition response headers) and
-// @kestra-io/hey-api-plugin's runtime.ts, which reads them to disambiguate. That header-reading
-// logic ships in @kestra-io/hey-api-plugin >= 0.3.0 (not yet released as of this pin at 0.2.0) —
-// re-pin the devDependency above once it is, no other change needed here.
+// TODO: re-pin @kestra-io/hey-api-plugin to >= 0.3.0 once released, for header-based disambiguation.
 const enterpriseFeature: EnterpriseFeatureConfig = {
     matchRoute: (method, path) => enterpriseOnlyRoutes[`${method} ${path}`],
     docsUrl: (feature) => `https://kestra.io/docs/enterprise-edition/${feature}`,
--- a/javascript/javascript-sdk/src/index.ts
+++ b/javascript/javascript-sdk/src/index.ts
@@ -1,12 +1,12 @@
 import { client } from "./openapi/client.gen"
 import { formDataBodySerializer } from "./openapi/client"
 import type { ResolvedRequestOptions } from "./openapi/client"
-import { createConfigureClient, EnterpriseFeatureError } from "@kestra-io/hey-api-plugin/runtime"
+import { createConfigureClient, EnterpriseFeatureError, SdkVersionMismatchError } from "@kestra-io/hey-api-plugin/runtime"
 import type { EnterpriseFeatureConfig } from "@kestra-io/hey-api-plugin/runtime"
 import { ENTERPRISE_ONLY_ROUTES_JSON } from "./openapi/sdk/enterpriseOnlyRoutes.gen"
 
 export * from "./openapi/index"
-export { EnterpriseFeatureError }
+export { EnterpriseFeatureError, SdkVersionMismatchError }
 
 declare global {
     interface Window {
@@ -188,7 +188,6 @@ const featureToSlugMap: Record<string, string> = {
     "login": "/auth/authentication",
 }
 
-// TODO: re-pin @kestra-io/hey-api-plugin to >= 0.3.0 once released, for header-based disambiguation.
 const enterpriseFeature: EnterpriseFeatureConfig = {
     matchRoute: (method, path) => enterpriseOnlyRoutes[`${method} ${path}`],
     docsUrl: (feature) => `https://kestra.io/docs/enterprise${featureToSlugMap[feature] ?? ""}`,
--- a/javascript/javascript-sdk/src/index.ts
+++ b/javascript/javascript-sdk/src/index.ts
@@ -145,13 +145,53 @@ const axiosLikeClient = {
 } as const
 
 // Route registry baked in at generation time from the EE spec's x-kestra tags; see
-// enterprise-only-routes-plugin.ts. docsUrl/contactSalesUrl are generic pending real links.
+// enterprise-only-routes-plugin.ts.
 const enterpriseOnlyRoutes: Record<string, { feature: string }> = JSON.parse(ENTERPRISE_ONLY_ROUTES_JSON)
+
+const featureToSlugMap: Record<string, string> = {
+    "audit-logs": "/governance/audit-logs",
+    "bindings": "/auth/rbac",
+    "auths": "/auth/authentication",
+    "banners": "/instance/announcements",
+    "cluster": "/instance/maintenance-mode",
+    "mcp": "",
+    "services": "",
+    "policies": "/governance",
+    "plugins": "/instance/versioned-plugins",
+    "worker-groups": "/scalability/worker-group",
+    "worker-queues": "",
+    "worker-auth": "/auth/credentials",
+    "kill-switches": "/instance/kill-switch",
+    "misc": "",
+    "invitations": "/auth/invitations",
+    "users": "/auth/rbac",
+    "notifications": "",
+    "service-account": "/auth/service-accounts",
+    "tenants": "/governance/tenants",
+    "ai": "",
+    "apps": "/scalability/apps",
+    "assets": "/governance/assets",
+    "blueprints": "/governance/custom-blueprints",
+    "blueprint-tags": "/governance/custom-blueprints",
+    "flows": "/governance",
+    "groups": "/auth/rbac",
+    "scim-groups": "/auth/scim",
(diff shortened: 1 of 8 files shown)


elevatebart:
CI is red on "Generate SDK" (`npm error 404` on `@kestra-io/hey-api-plugin@0.3.0`) — expected, not a bug in this PR. The pin was bumped ahead of the companion `kestra` PR that publishes `0.3.0`; `npm install` can't resolve it until that release exists. Will recheck once it's published.

---

fdelbrayelle:
I advise you to put @jymaire and @Malaydewangan09 for the next 3 weeks as reviewers since I'll be off

javascript/javascript-sdk/src/index.ts
```diff
@@ -141,7 +144,57 @@ const axiosLikeClient = {
...
+const enterpriseOnlyRoutes: Record<string, { feature: string }> = JSON.parse(ENTERPRISE_ONLY_ROUTES_JSON)
+
+const featureToSlugMap: Record<string, string> = {
+    "audit-logs": "/governance/audit-logs",
+    "bindings": "/auth/rbac",
+    "auths": "/auth/authentication",
+    "banners": "/instance/announcements",
+    "cluster": "/instance/maintenance-mode",
+    "mcp": "",
+    "services": "",
```
jymaire, line 159: some missing slugs, is it normal?

jymaire:
Review notes below. The design is sound — deriving the registry from the spec at generation time is the right call versus a hand-maintained list, and the `x-kestra` tags are already present on this branch (321 EE-tagged operations across 38 tags), so this isn't waiting on the spec.

Skipping the dependency/lockfile blockers here since you already documented them in the description. Everything below is static analysis — CI dies at `npm ci` on the `v0.3.0` 404, so nothing in this PR, including the new test, has been exercised yet.

## Correctness

### 1. `credentials` is missing from `featureToSlugMap`

I extracted every EE-tagged operation from `kestra-ee.yml` on this branch and applied the plugin's `kebabCase` to the first tag. 38 features result; the map has 37 keys. The gap:

```
IN SPEC, NOT IN MAP: ['credentials']   → 13 operations
    GET/POST   /api/v1/{tenant}/credentials
    GET/PUT/DELETE /api/v1/{tenant}/credentials/{id}
    …
```

Those 13 routes fall through `?? ""` to the generic `https://kestra.io/docs/enterprise`. Suspicious detail: the map has `"worker-auth": "/auth/credentials"` — the credentials docs slug looks like it landed on the wrong key. No map key is dead, so this is the only drift.

### 2. 82% of EE routes are templated, and none are tested

265 of the 321 registry keys contain path params (`{tenant}`, `{id}`, …), e.g. `GET /api/v1/{tenant}/credentials/{id}`. Lookup is exact-string:

```ts
matchRoute: (method, path) => enterpriseOnlyRoutes[`${method} ${path}`]
```

So this only works if the runtime hands `matchRoute` the **un-substituted** path. If it passes the resolved URL (or lowercases the method, or prefixes `baseUrl`), the feature silently no-ops for the large majority of routes **while still passing the one test in this PR**.

The test comment says it picks a route "so `matchRoute` gets a templated path", but `GET /api/v1/auditlogs/search` has no path params — it's one of the 56 non-templated routes. Worth adding a case over a `{tenant}`-scoped route.

### 3. The genuine-404 case is untested, and it's the one that can misfire

The test stubs a 404 with no `X-Kestra-Edition` / `X-Kestra-Entity` headers — i.e. the legacy-server fallback branch. On a real EE instance, `GET /api/v1/{tenant}/credentials/{missing-id}` is *also* a 404 on a registry-known route. If the runtime doesn't prefer the headers over the registry, the SDK tells an existing EE customer to go buy Enterprise.

Those headers are the whole point of the companion `kestra` PR, so presumably the runtime does prefer them — but nothing here proves it. Suggest the inverse assertion: 404 **with** `X-Kestra-Edition: ee` → plain error, not `EnterpriseFeatureError`.

### 4. Silent-drift risk in the plugin

If `x-kestra` is ever renamed or dropped upstream, `routes` becomes `{}`, generation succeeds, the build succeeds, `tsc` passes — and the feature is just gone, with no signal. Cheap guard: fail (or at least `console.warn`) in the handler when zero EE routes are collected.

Pairs well with a unit test asserting

```ts
new Set(Object.values(routes).map(r => r.feature)) ⊆ new Set(Object.keys(featureToSlugMap))
```

which would have caught finding 1 automatically, and keeps catching it as kestra-ee adds tags.

## Code quality

- **Feature keys are derived from display tags.** `kebabCase(operation.tags?.[0])` means a cosmetic rename in kestra-ee (`"Audit Logs"` → `"Audit log"`) silently breaks the slug lookup with no error. Worth considering putting the slug in the spec itself (`x-kestra: {edition: ee, docs: "/governance/audit-logs"}`) so the two can't drift — that also deletes the 37-line map from `src/index.ts`.
- **`?? ""` hides unknown features.** Because known-but-unmapped features legitimately map to `""`, an *unknown* feature is indistinguishable from a deliberate one. `Object.hasOwn` plus the exhaustiveness test above fixes both.
- **`operation: any`** discards the `IR.OperationObject` type. Prefer typing the param and narrowing only the `x-kestra` read, which you already do correctly via the inline cast.
- **`kebabCase` duplicates** the near-identical transform in `tsdown.config.ts:9`.
- **Test hygiene:** `vi.unstubAllGlobals()` sits at the end of the test body, so a failed assertion leaves `fetch` stubbed. Move it to `afterEach`. Vitest's per-file isolation contains the blast radius today, but this file will grow.
- **`beforeEach` drops auth.** `_setup.ts:12` configures the client with credentials; the new `configureClient({ baseUrl })` replaces it without `auth`. Correct for this stubbed test and fine under file isolation — just fragile if a live-backend test is ever added to the same file.
- The `"./enterprise-only-routes"` export in `package.json` lines up with `tsdown.config.ts`'s auto-derived entries (`enterpriseOnlyRoutes.gen.ts` → `enterprise-only-routes`), so that's correct — it does make a raw JSON string part of the public API surface, though. Intentional?

## Minor

- `JSON.parse` of the ~320-entry registry runs at import of `src/index.ts`, so every consumer pays it, EE or not. Small, and `JSON.parse`-of-a-string-literal is genuinely the fast option here — just noting it's eager.
- `$.literal(JSON.stringify(routes))` relies on the codegen escaping the string correctly; worth eyeballing the emitted `.gen.ts` once (sounds like you have).
- The new `.gen.ts` lands under `src/openapi/sdk/**`, which is in the coverage `include` glob. It has no functions, so the `functions: 75` threshold is unaffected.
- `README_JAVASCRIPT_SDK.md` isn't updated for the newly exported error types.
- No `### ⚠️ Signature change` note needed — the changes are additive and `configureClient`'s consumer-facing signature is unchanged. Correct per `AGENTS.md`.

javascript/javascript-sdk/src/index.ts
elevatebart: Dug into this one — it's not actually missing. `configurations/kestra-openapi-sdk-customizer.json` has `"tagsToSkip": ["Credentials"]`, so that tag's 13 operations never reach the JS SDK generator at all; `credentials` can never appear in the registry. `worker-auth: "/auth/credentials"` was already the correct mapping (for the separate, non-skipped `Worker Auth` tag) — I left it as-is. Full explanation in my reply on the main review thread.

---

elevatebart:
Thanks for the deep dive — addressed in the latest commit:

**1. `credentials` slug** — investigated, and it's not actually a gap: `configurations/kestra-openapi-sdk-customizer.json` has `"tagsToSkip": ["Credentials"]`, so that tag's 13 operations are filtered out before `openapi-ts` ever runs. `credentials` can never appear in the generated registry, and `worker-auth: "/auth/credentials"` was already the correct mapping for the separate `Worker Auth` tag. No map change made — left as-is (replied on the inline thread too).

**2. Templated-path coverage** — added a second case in `useClientEnterpriseFeature.spec.ts` using `invitation()` (`GET /api/v1/{tenant}/invitations/{id}`), confirming `matchRoute` still gets the templated path for a `{tenant}`-scoped route.

**3. Headers-take-precedence-over-registry case** — this is exactly what the re-added `runtime.spec.ts` in the companion `kestra` PR covers (entity-header-present → never `EnterpriseFeatureError`, regardless of edition). That's the right place for it since it's testing the shared interceptor logic itself, not this SDK's generated registry.

**4. Silent-drift guard** — added a `console.warn` in the plugin handler when zero EE routes are collected, plus a new `featureToSlugMap.spec.ts` asserting every registry feature has a map entry (exports `featureToSlugMap` from `index.ts` for that purpose).

**5. `operation: any`** — removed; relies on inference now, only the `x-kestra` read is cast (`as unknown as {...}`).

**6. Test hygiene** — moved `vi.unstubAllGlobals()` to `afterEach`.

**7. README** — added a short section documenting `EnterpriseFeatureError`/`SdkVersionMismatchError`.

**Skipped, by design:**
- De-duping `kebabCase` with `tsdown.config.ts`'s inline version — cosmetic, touches build config for no behavior change.
- Moving the slug into the spec itself (`x-kestra.docs`) — good idea, but that's kestra-ee's tagging script, out of scope for this PR.
- `enterprise-only-routes` as a public export — yes, intentional; it's a plain generated JSON string, cheap to expose, and needed internally by `index.ts`.

Still blocked on CI (`@kestra-io/hey-api-plugin@0.3.0` 404) until the companion `kestra` PR publishes that release, as already noted on the thread.

---

jymaire (approved):
Approving — the review points are addressed, and I verified them rather than taking them on faith: the `credentials` explanation holds (`"tagsToSkip": ["Credentials"]` is confirmed in `configurations/kestra-openapi-sdk-customizer.json`, so that tag can never reach the generated registry), the templated-path case is there via `invitation()`, the zero-routes drift guard and exhaustiveness test are in, and the headers-take-precedence case really is covered by `runtime.spec.ts` in the companion `kestra#17622`.

Two things still gate the actual merge, so please don't squash on green-looking approval alone:

1. `kestra#17622` has to merge and publish `@kestra-io/hey-api-plugin@0.3.0` first — the pinned tarball 404s today (`hey-api-plugin` is still at `v0.2.0`).
2. `package-lock.json` still resolves `v0.2.0` (with 0.2.0's integrity hash) while `package.json` pins `v0.3.0`. `npm ci` fails on that mismatch, so this needs a lockfile regeneration commit after 0.3.0 ships — not just a CI re-run.

Non-blocking nits, take or leave:

- `featureToSlugMap.spec.ts` only asserts key *presence*, so the nine features mapped to `""` pass silently and `docsUrl` returns a bare `https://kestra.io/docs/enterprise` for them. If that's intentional, an explicit "no dedicated docs page" allowlist would make the intent self-documenting and give the test something real to guard.
- `featureToSlugMap` lands in the published API surface for the sake of one test. Exporting the generated registry subpath is fine; this one is test-only plumbing.

elevatebart:
Thanks for verifying rather than taking it on faith — appreciated.

Confirmed both blockers, agreed on not merging yet:

1. `kestra#17622` publishing `@kestra-io/hey-api-plugin@0.3.0` is a hard prerequisite, tracked there.
2. Checked `javascript/package-lock.json` — you're right, it still has the `0.2.0` resolved entry/integrity hash while `package.json` pins `0.3.0`. Can't regenerate it correctly against a tarball that 404s today, so I'll run `npm install` (not `ci`) to refresh the lockfile as soon as `0.3.0` is published, as its own commit, and ping this thread.

On the nits — leaving both as-is, but for the record:
- The `""` slugs in `featureToSlugMap` are an intentional "no dedicated docs page yet" fallback, consistent with how the rest of the pre-existing map already handles undocumented features — not something this test is meant to catch. Open to a follow-up allowlist if that assumption ever needs to be more explicit.
- Agreed `featureToSlugMap`'s export is test-only plumbing, not consumer API. Keeping it since splitting it into a separate module just to avoid one extra named export felt like more churn than the nit warranted, but flagging I considered it.

---

florian-create:
Coming in from the growth side, non blocking.

One idea. That link is the only thing we get to say to someone at the exact moment they run into an Enterprise boundary. Sent bare, if they follow it the visit looks like any other visit to our docs, so we never find out whether the message helped, or which feature sent them there.

The app already tags its own links to the docs for that reason, e.g.

```
/docs/workflow-components/inputs?utm_source=app&utm_medium=referral&utm_campaign=editor-flow-doc
```

Would you be up for the SDK doing the same, with the feature key at the end?

```
<docsUrl>?utm_source=sdk&utm_medium=referral&utm_campaign=ee-feature-error&utm_content=audit-logs
```

The per-feature slugs you already built do most of the work here. The feature key in `utm_content` mainly helps for the ones that currently fall back to the generic `/docs/enterprise`, where the URL alone can't tell them apart.

Same convention as the app, nothing new to decide, nothing to maintain on your side afterwards. It just means we can tell whether these links do anything.

elevatebart:
@florian-create good call, done in 3d9568a — you're right that this is the one moment we get to say anything, and sending the link bare throws away the only signal that would tell us whether it worked.

Followed the app's existing convention exactly (`utm_source=app&utm_medium=referral&utm_campaign=…&utm_content=…`, as in `BasicAuthLogin.vue` and `useDocsLink.ts`), swapping in `utm_source=sdk`:

```
https://kestra.io/docs/enterprise/governance/audit-logs?utm_source=sdk&utm_medium=referral&utm_campaign=ee-feature-error&utm_content=audit-logs
```

Two notes on what I did beyond the literal ask:

- **Tagged `contactSalesUrl` too**, not just `docsUrl`. `EnterpriseFeatureError` carries both, and the app already tags its `kestra.io/demo` links (`DemoButtons.vue`, `Empty.vue`), so leaving the sales link as the one untagged outbound link seemed like the obvious next request. Same campaign, same `utm_content`. Easy to drop back to docs-only if you'd rather keep the demo campaign space clean.
- **`utm_content` is on every feature**, not only the ones falling back to the generic `/docs/enterprise`. Costs nothing and means you can segment by feature without reverse-mapping slugs.

Test coverage: `useClientEnterpriseFeature.spec.ts` now asserts both full URLs, so a future slug edit that drops the tagging fails the suite rather than going quiet. Also added a line to `README_JAVASCRIPT_SDK.md` so consumers rendering these URLs know the query string is there and can strip it if they don't want to pass it through.

Merge is still gated on the two things from @jymaire's approval — `kestra#17622` publishing `@kestra-io/hey-api-plugin@0.3.0` (still 404s as of today), then the lockfile regeneration commit. This change is independent of both, but it hasn't been executed either: CI can't get past `npm ci`, so the new assertions are unrun. I'll confirm them green with the rest once `0.3.0` lands.

---

jymaire:
Hello @elevatebart, is this PR still needed?

elevatebart:
`Test Javascript SDK` on `e479e108` reported 82 failures. One was this PR's and is fixed in `cdcfb6ca`; the other 81 are not.

**Fixed here:** `featureToSlugMap.spec.ts` — the generated EE-only route registry has grown `cases`, `dashboards` and `promote` since this branch was written, and the exhaustiveness spec fails on any feature the slug map does not cover. Added `/governance/cases` and `/governance/promote`; dashboards has no page under `/docs/enterprise`, so it takes the empty-slug fallback already used for `mcp`, `services` and `quotas`.

**Not this PR's:** the remaining 81 are all `KestraProblemError: Access denied` (a 403) across `AppsApi`, `AssetsApi`, `BlueprintsApi`, `CasesApi`, `DashboardsApi`, `PoliciesApi` and three more EE-surface specs. The same 81 failures, in the same 8 spec files, hit PR #455 on an unrelated branch ([run 35595701093](https://github.com/kestra-io/client-sdk/actions/runs/35595701093)). The last green run on `main` was 10:05 UTC today, so the test instance lost its EE authorization somewhere after that. This PR changes 404 handling only and cannot produce a 403; no fix for it exists in this repo yet, so there is nothing to port.

---

elevatebart:
Re-ran after the EE image was published: **83 failures → 1**. The `Access denied` wave from my earlier comment is gone.

The one remaining failure is not this PR's either, and #417 fixes it:

`MiscApi.spec.ts` → `generate: returns the usage report as a file, defaulting 'from' to the current month`. The reported `AssertionError: expected 'undefined' to be 'number'` is misleading: `expect(result).toBeInstanceOf(Blob)` sits inside the test's `try`, so its own AssertionError is caught and handed to `expectHttpStatus`, which then asserts on a `status` an AssertionError doesn't have. The real failure is `generate()` not returning a `Blob`.

Cause is the spec. On `main`, `/api/v1/{tenant}/stats/generate-reports` declares its `application/octet-stream` 200 as `type: string, format: byte`, so hey-api generates a string response. #417's synced spec declares `format: binary`, which generates a `Blob` and satisfies the assertion. The test arrived unskipped in `1857e0dd` ("the usage report ... comes back as a file") ahead of that spec, so it is red on `main` as well; `MiscApi.spec.ts` here is byte-identical to `main`'s.

Nothing to port: the fix *is* #417's diff. Holding this PR until it merges, as agreed.

---

PR #370, opened 2026-07-24
feat: hand-write Java SDK bindings for Cases API

brian-mulier-p:
- Adds `CasesApi` + Cases model classes (`Case`, `SlaConfig`, `Subjects`, `CaseSeverity`, `CaseStatus`, `CasesControllerCaseFromTaskRequest`) following the existing hand-written generated-code idiom
- Wires `CasesApi` into `KestraClient`
- `SlaConfig` exposes `acknowledgement` + `resolution` (both counted from case creation)

Companion to kestra-ee PR #9383 (Cases feature) and a plugin-kestra PR adding the `CreateCase` task.


fdelbrayelle:
## QA against a real Kestra EE instance

Ran the new bindings against a local EE instance built from the `v2.0.0-rc11` tag rather than the CI Docker image.

**Environment**
- Kestra EE `v2.0.0-rc11` built from the tag (EE `8823a5c72`, OSS `751ad1e905`), Postgres backend on a fresh database, API on `http://localhost:9901`
- Tenant `main`, authenticated as super-admin with an API token
- SDK built from this branch (`54f2b5a`) and consumed as a published artifact

`CasesController.CaseFromTaskRequest` is identical between `v2.0.0-rc11` and `develop`, so this exercises the released contract rather than a moving branch.

### `POST /api/v1/{tenant}/cases/from-task` — via `CasesApi.createFromTask`

| # | Scenario | Expected | Result |
|---|---|---|---|
| 1 | Minimal request (`namespace`, `title`, `flowNamespace`, `flowId`, `taskId`) | `created=true`, `caseId` returned | ✅ |
| 2 | All optional fields set: `description`, `severity=CRITICAL`, `status=INVESTIGATING`, `sla{acknowledgement=PT1H, resolution=PT8H}`, `assignees`, `watchers`, `labels`, `actions` | every field stored verbatim, retrievable on `GET /cases/{id}` | ✅ |
| 3 | `status` omitted | server defaults to `OPEN` (nullable enum is omitted from the payload, not sent as `null`) | ✅ |
| 4 | `labels` sent | stored as `customFields` server-side | ✅ |
| 5 | `origin` after creation | `{flowNamespace, flowId, taskId}` match the request | ✅ |
| 6 | Same origin twice with `linkMatchingExecutions=true` | second call returns `created=false` with the first `caseId` | ✅ |
| 7 | Same origin twice with `linkMatchingExecutions=false` | two distinct cases created | ✅ |
| 8 | Explicit `caseId` | attaches instead of creating: `created=false`, same id | ✅ |
| 9 | Unknown `caseId` | `404` surfaced as `ApiException` | ✅ |
| 10 | `title` omitted | `400` surfaced as `ApiException` (server-side rule: title required when creating) | ✅ |
| 11 | Required `flowId` omitted | `4xx` surfaced as `ApiException` | ✅ |

Serialization of `CaseSeverity`, `CaseStatus`, `SlaConfig`, `Subjects`, `CaseAction` and `Label` was checked by reading the created case back from the API, not just by asserting on the response of the create call.

### `GET /api/v1/{tenant}/flows/source` — via the reworked `FlowsApi.searchFlowsBySourceCode`

| # | Scenario | Expected | Result |
|---|---|---|---|
| 12 | `caseSensitive=false` vs `true` on a case-mismatched query | 1 hit vs 0 hits | ✅ |
| 13 | `regex=true` with `regexm.rker` | matches the flow containing `regexmarker` | ✅ |
| 14 | `scope=TASKS` vs `scope=ALL` for a marker in the flow description | 0 hits vs 1 hit | ✅ |
| 15 | Response mapping | `SourceSearchResult.id/namespace/matches` populated, `SourceMatch.line/snippet` populated | ✅ |

All four new query params reach the server and change the result set, so the enum is serialized by value (`ALL`/`TASKS`/…) and the booleans are not dropped.

### Not covered

- `assetIds` — needs a pre-existing asset in the instance
- ACL paths (`CASE.CREATE` / `CASE.UPDATE` denials, cross-namespace dedup-attach) — everything ran as super-admin
- Concurrent `from-task` calls on the same origin — the endpoint documents itself as check-then-act, so a race is expected there by design

### Verdict

No functional defect found in the bindings. Every field of `CasesController.CaseFromTaskRequest` round-trips, error responses surface as `ApiException` with the right status code, and the `/flows/source` rework behaves against the live EE API on `v2.0.0-rc11`.

Remaining review comments on this PR are non-functional (missing `### ⚠️ Signature change` note for `searchFlowsBySourceCode`, stale `docs/` + `README` entries for the deleted `SearchResultFlow`, and no docs for the new `CasesApi`).

fdelbrayelle (approved):

PR #372, opened 2026-07-28
fix(javascript): align the Policies and Quotas specs with the YAML-source policy API

jymaire:
Fixes the `javascript SDK vs latest Kestra` leg of the daily freshness check, red since [run 30333124607](https://github.com/kestra-io/client-sdk/actions/runs/30333124607) (15 of 16 `PoliciesApi` tests failing with a bare `403 Forbidden`).

## Root cause

kestra-ee [`fe7140415`](https://github.com/kestra-io/kestra-ee/commit/fe7140415317aa690fbafe5a62e801ff55a220a9) (*feat(policies): populate and round-trip the policy source*) switched every policy create/update from a JSON body to a YAML source:

```java
- @Post(uri = "/{tenant}/policies")                              @Body @Valid ApiPolicyRequest request
+ @Post(uri = "/{tenant}/policies", consumes = APPLICATION_YAML) @Body @NotNull String source
```

`main`'s committed `kestra-ee.yml` predates that, so the regenerated SDK posts `application/json` to a route that only consumes `application/x-yaml`. No route matches, and EE's security filter rejects the unmatched `/api` path before routing — which is why it surfaces as `403` rather than `404`/`415`.

Every failure lands on a `create*Policy`/`update*Policy` call; `importPolicies` is the only test that stayed green because it is `multipart/form-data` and was left untouched. That pattern rules out a permission or license-gate cause.

## Changes

- `kestra-ee.yml` refreshed up to kestra-ee@`1e1d45ca` (the five `ci: auto-update` commits, same content as #369).
- `PoliciesApi.spec.ts` — policies are now created and updated from their YAML source; the spec builds the document with a `policySource()` helper instead of a structured `policyRequest()` object.
- `QuotasApi.spec.ts` — follows the `search` → `searchQuotaLimits` operationId rename.

## ⚠️ Signature change

Generated JS SDK, driven by the spec refresh — all breaking, no back-compat aliases (they mirror upstream route/operation renames):

```
Policies.createInstancePolicy({ ...PolicyRequest })   → Policies.createInstancePolicy({ body: string /* YAML */ })
Policies.updateInstancePolicy({ ...PolicyRequest })   → Policies.updateInstancePolicy({ id, body: string })
Policies.createTenantPolicy({ ...PolicyRequest })     → Policies.createTenantPolicy({ body: string })
Policies.updateTenantPolicy({ ...PolicyRequest })     → Policies.updateTenantPolicy({ id, body: string })
Policies.createNamespacePolicy({ ...PolicyRequest })  → Policies.createNamespacePolicy({ namespace, body: string })
Policies.updateNamespacePolicy({ ...PolicyRequest })  → Policies.updateNamespacePolicy({ namespace, id, body: string })
Quotas.search(...)                                    → Quotas.searchQuotaLimits(...)
Dashboards.exportChartToCsv(...)                       → Dashboards.exportChart({ ..., format?: 'CSV' | 'ION' })
Dashboards.exportDashboardChartDataToCSV(...)          → Dashboards.exportDashboardChart({ ..., format?: 'CSV' | 'ION' })
```

New operations added by the same refresh: `validateInstancePolicy`, `validateTenantPolicy`, `validateNamespacePolicy`, `getDefaultDashboardDefinitions`, `resetQuotaLimit`.

Downstream consumers (plugin-kestra) should note the `Policies.*` and `Quotas.search` renames.

## Relationship to other PRs

- Supersedes #369 — same `kestra-ee.yml` content, plus the test realignment that #369 alone would leave red.
- Overlaps #323 on the two dashboards export paths in `kestra-ee.yml`: #323 hand-edits them, this PR picks them up from the full spec regen. Whichever lands second will need the other's hunk dropped. No JS test exercises the dashboards export, so the JS leg is unaffected either way; #323 still owns the Java/Python/Go method renames.

### DIFF
--- a/javascript/test_javascript_sdk/PoliciesApi.spec.ts
+++ b/javascript/test_javascript_sdk/PoliciesApi.spec.ts
@@ -5,27 +5,30 @@ import { tenantId } from './_setup.js';
 import * as Policies from '@kestra-io/kestra-sdk/policies';
 import * as Namespaces from '@kestra-io/kestra-sdk/namespaces';
 import * as Flows from '@kestra-io/kestra-sdk/flows';
-import type { PolicyRequest, Rule } from '@kestra-io/kestra-sdk';
 
+// Policies are created and updated from their YAML source (`application/x-yaml`): the API
+// parses it, stamps the scope/tenant/namespace from the URL and stores the source alongside
+// the parsed model, so the body is a plain string rather than a structured request object.
+//
 // `io.kestra.plugin.ee.rules.Require` is a built-in FLOW-scoped governance rule: it reports
 // a violation when the listed properties are missing from a flow after mutation.
-function requireTimeoutRule(): Rule {
-    return {
-        type: 'io.kestra.plugin.ee.rules.Require',
-        on: 'FLOW',
-        action: 'WARN',
-        errorMessage: 'timeout is required',
-        properties: ['timeout'],
-    } as Rule;
-}
-
-function policyRequest(id: string): PolicyRequest {
-    return {
+function policySource(id: string, overrides: Record<string, string> = {}): string {
+    const fields: Record<string, string> = {
         id,
         displayName: `Test policy ${id}`,
         enforcement: 'EVALUATE',
-        rules: [requireTimeoutRule()],
+        ...overrides,
     };
+    return [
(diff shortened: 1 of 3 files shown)


brian-mulier-p (approved):
LGTM

PR #375, opened 2026-07-30
fix(tests): stay within the server's 1000 max page size

jymaire:
## Why

The nightly [SDK freshness check](https://github.com/kestra-io/client-sdk/actions/runs/30517898385/job/90791707577) went red on the `javascript` leg. Two of the failures were a real upstream breaking change:

kestra-io/kestra `7177bc9c04` and kestra-ee `d5af3252f` — *"enforce a max page size of 1000 on list/search endpoints"* — added `PageableUtils.MAX_PAGE_SIZE = 1000` at the choke point every OSS + EE list/search endpoint routes through, plus `@Max(1000)` on each `size` query param. Any `size` above the cap is now rejected with a **422** instead of being served:

```
422 Invalid entity: searchFlowsBySourceCode.size: must be less than or equal to 1000
```

Four JS tests asked for `size: 10000` and broke. Go/Java/Python stayed green by luck, not by coverage — none of them happened to ask for more than 1000.

## What

**1. Pin the boundary in every SDK's suite.** Each test asserts both sides — the cap itself must be served (checking the created flow actually comes back, not just that a response is defined), one above it must 422:

| SDK | Test | Constant |
|---|---|---|
| JS | `FlowsApi.spec.ts` → `search_flows_by_source_code: rejects a page size above the cap` | `_utils.MAX_PAGE_SIZE` |
| Go | `api_flows_test.go` → `searchFlowsBySourceCodePageSizeCapTest` | `common_test_setup.go` |
| Java | `FlowsApiTest#searchFlows_pageSizeCap` | `TestUtils.MAX_PAGE_SIZE` |
| Python | `test_phase4_filters_pagination_tenant.py` → `..._at_max_page_size`, `..._above_max_page_size_rejected` | `test_helpers.MAX_PAGE_SIZE` |

The four `size: 10000` call sites themselves are fixed on `main` by #378, which landed while this PR was open and narrows them to `size: 5` plus an `EQUALS` filter on the name — a better fix than clamping to the cap, since it asserts against a targeted result set. This PR keeps that and adds the boundary coverage #378 doesn't have. A repo-wide sweep found no other site above 1000 in any SDK.

**2. No more silently-wrong execution state.** `awaitExecution` returned whatever state it last saw once its timeout expired. That is what turned a timing problem into a fake API break: the bulk-pause tests reported `400 invalid bulk pause` because their executions had already completed, which reads as an upstream break rather than a slow fixture. It now throws with expected vs. actual state.

Paired with that, since a timeout is now a hard failure:
- the 1500/2000ms budgets that relied on the lenient behaviour go to 5000ms;
- `SLEEP_CONCURRENCY_FLOW` sleeps `PT10S` instead of `PT2S`, so the RUNNING / QUEUED state a test sets up is still true when it acts on it;
- `pause_executions_by_ids` and `pause_executions_by_query` assert their non-matching execution is **still RUNNING** instead of waiting for it to succeed — "it finished" would also hold for an execution that was never a candidate for pausing.

**3. A fixture that could never succeed.** The Logs / Metrics / Outputs / Dashboards suites each polled for `SUCCESS || FAILED` and fell through silently on timeout. They now share a `waitForExecutionSuccess` helper that throws on a terminal failure and on timeout — which immediately turned 16 tests red, because all four built their fixture execution from `simple_flow.yml`, and that flow can never succeed:

```
io.kestra.core.exceptions.IllegalVariableEvaluationException:
Unable to find `vars` used in the expression `The variables we used are
{{ vars.first }} and {{ render(vars.second) }}. ...`
```

Its only task renders `{{ vars.first }}`, `{{ render(vars.second) }}` and `{{ inputs.user }}` while the flow declares no `variables:` and no `inputs:`, so every execution of it has always ended `FAILED`. The old lenient poll accepted that and the suites asserted against an execution that had never run a task — MetricsApi's `toBeDefined` checks in particular were passing vacuously.

Fixed with `getExecutableFlowAndId()`: the same flow with the declarations its message references, keeping the `my_task_1_id` task id that the per-task metric lookups pass. `simple_flow.yml` is left as it is for the flow CRUD tests, which never execute it — the same split the Go suite already makes between `simple_flow.yml` and its own `LOG_FLOW`, which is why Go never hit this.

## Verification

CI is green on all four SDKs. The JS leg went from 42 failures across 8 files to **47 files passed / 1 skipped / 0 failures**, and got *faster* than `main`:

| | Before the fixture fix | After | `main` |
|---|---|---|---|
| Wall clock | 388s | **161s** | 178s |
| Test time | 1132s | **454s** | 503s |
| Failures | 42 | **0** | 0 |

25 of those 42 failures were 5s timeouts in specs this PR never touches (`KvApi` 10, `BlueprintsApi` 7, `AssetsApi` 2, `ExecutionsApi` read-only tail 6). They were starvation, not broken tests: 16 fixture tests × 4 attempts (`retry: 3`) × a 10s wait each ≈ 640s of the run's 1132s test time, every attempt creating another flow and another failing execution against one shared container at `maxWorkers: 3`. Removing the failing fixture cleared all of them with no change to those specs — `KvApi` runs 11/11 in 320ms locally, `BlueprintsApi` 13/13 in 470ms.

The fixture change was also checked directly against a local Kestra EE before pushing: the old flow ends `FAILED`, the new one `SUCCESS`, and LogsApi (6/6), MetricsApi (6/6), OutputsApi (2/2) and ExecutionsApi (56 pass / 1 skipped) all pass.

No public SDK method signature changed — this PR is tests and test constants only.

## Not in this PR

1. `test-utils/flows/simple_flow.yml` is still un-executable for whoever reaches for it next. Fixing it at the source would help all four SDKs, but it is shared with the Java / Python / Go flow-CRUD tests, so it deserves its own PR rather than riding along here.
2. The freshness run that started this was ~6x slower end to end than the previous day (42s → 166s wall, 80s → 470s test time). Worth chasing upstream on its own.

### DIFF
--- a/.github/scripts/coverage-comment.mjs
+++ b/.github/scripts/coverage-comment.mjs
@@ -1,15 +1,28 @@
 #!/usr/bin/env node
 /**
- * Reads a vitest/istanbul coverage-final.json and posts (or updates) a sticky
- * comment on the pull request listing every uncovered function in openapi/sdk/,
- * and optionally lists failing tests from a vitest JSON report.
+ * Posts (or updates) a sticky comment on the pull request with the JavaScript
+ * SDK test status:
  *
- * Usage: node coverage-comment.mjs <path-to-coverage-final.json> [path-to-test-results.json]
+ *  - failing tests, grouped by spec file, each one linking to its source line
+ *    on GitHub and showing the recorded error, so a red build is readable from
+ *    the PR without digging through the job logs;
+ *  - every uncovered function in `openapi/sdk/`.
  *
- * Required env vars:
- *   GITHUB_REPOSITORY  – e.g. "owner/repo"
- *   PR_NUMBER          – pull-request number
- *   GH_TOKEN           – GitHub token with pull-requests:write
+ * Failure data comes from `coverage/test-failures.json` (written by
+ * test_javascript_sdk/failureReporter.ts), falling back to vitest's built-in
+ * json report when that file is absent.
+ *
+ * Usage: node coverage-comment.mjs <coverage-final.json> [test-results.json] [test-failures.json]
+ *
+ * Env vars:
+ *   GITHUB_REPOSITORY  – e.g. "owner/repo"       (required to post)
+ *   PR_NUMBER          – pull-request number     (required to post)
+ *   GH_TOKEN           – token with pull-requests:write
+ *   COMMIT_SHA         – sha the spec links point at (defaults to GITHUB_SHA)
+ *   RUN_URL            – link to the workflow run, shown next to the failures
+ *   TESTS_OUTCOME      – outcome of the test step ("success" / "failure" / …).
+ *                        Used to still report something when the run died
+ *                        before writing any report.
  */
 
 import { readFileSync, writeFileSync, existsSync } from "fs";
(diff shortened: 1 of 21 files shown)


Malaydewangan09 (approved):
LGTM, some conflicts to resolve

PR #376, opened 2026-07-30
ci: enhance JS SDK test reporting with detailed failure info

elevatebart:

This PR enhances the JavaScript SDK test reporting in pull requests by:

1. **Detailed test failure reporting**: The PR comment now displays failing tests grouped by spec file, with:
   - Direct links to the failing test source code on GitHub (at the correct line number)
   - Full error messages (truncated intelligently to stay readable)
   - Test hierarchy (suite names) for context
   - Automatic expansion for small failure lists, collapsible details for larger ones

2. **Graceful handling of test infrastructure failures**: When the test run fails before producing a report (e.g., syntax errors, import failures, setup issues), the comment now reports this instead of silently showing only coverage data.

3. **Improved script robustness**:
   - Accepts both vitest and jest JSON report formats
   - Handles suite-level failures (collection errors) in addition to individual test failures
   - Respects GitHub's 65KB comment size limit with graceful truncation
   - Configurable via environment variables for flexibility in CI/CD pipelines

4. **Better developer experience**:
   - Developers can see exactly what failed without digging through logs
   - Links point directly to the failing test source
   - Includes reproduction instructions in the comment
   - Workflow run URL is linked for full logs when needed

### Changes summary

- **`.github/scripts/coverage-comment.mjs`**: Completely refactored to parse and display test failures with rich formatting, while maintaining existing coverage reporting
- **`.github/workflows/javascript-sdk.yml`**: Updated to pass test result metadata (commit SHA, run URL, test outcome) to the comment script; comment now runs even on test failure
- **`javascript/test_javascript_sdk/vitest.config.ts`**: Enabled `includeTaskLocation` in the vitest JSON reporter to capture test line numbers for source linking

---

The changes are validated by:
- Existing CI/CD pipeline that runs the test suite
- The script gracefully degrades when coverage or test result files are missing
- Environment variable handling is defensive (uses defaults, checks for required values before posting)
- Comment body size is clamped to GitHub's limits with truncation warnings

No manual QA needed — the script will be exercised on every PR that runs the JavaScript SDK tests.

https://claude.ai/code/session_01HfvTcMTqxDMZiTsZXmbEm2

### DIFF
--- a/.github/scripts/coverage-comment.mjs
+++ b/.github/scripts/coverage-comment.mjs
@@ -8,7 +8,11 @@
  *    the PR without digging through the job logs;
  *  - every uncovered function in `openapi/sdk/`.
  *
- * Usage: node coverage-comment.mjs <path-to-coverage-final.json> [path-to-test-results.json]
+ * Failure data comes from `coverage/test-failures.json` (written by
+ * test_javascript_sdk/failureReporter.ts), falling back to vitest's built-in
+ * json report when that file is absent.
+ *
+ * Usage: node coverage-comment.mjs <coverage-final.json> [test-results.json] [test-failures.json]
  *
  * Env vars:
  *   GITHUB_REPOSITORY  – e.g. "owner/repo"       (required to post)
@@ -42,9 +46,12 @@ const coveragePathArg =
     process.argv[2] ?? "javascript/coverage/coverage-final.json";
 const testResultsPathArg =
     process.argv[3] ?? "javascript/coverage/test-results.json";
+const failuresPathArg =
+    process.argv[4] ?? "javascript/coverage/test-failures.json";
 const repoRoot = join(import.meta.dirname, "..", "..");
 const coveragePath = join(repoRoot, coveragePathArg);
 const testResultsPath = join(repoRoot, testResultsPathArg);
+const failuresPath = join(repoRoot, failuresPathArg);
 
 const {
     GITHUB_REPOSITORY: repo,
@@ -112,10 +119,50 @@ function formatError(rawMessages) {
  *          line: number | null; error: string }[]}
  */
 const failingTests = [];
+/** @type {string[]} */
+const unhandledErrors = [];
 let testResultsFound = false;
 let testRunSucceeded = null;
 
-if (existsSync(testResultsPath)) {
+// Preferred source: our own reporter, which keeps the real error message for
+// timed-out tests (vitest's json report replaces it with an internal sentinel).
+if (existsSync(failuresPath)) {
+    try {
+        const report = JSON.parse(readFileSync(failuresPath, "utf8"));
+        testResultsFound = true;
+
+        for (const failure of report.failures ?? []) {
+            failingTests.push({
+                file: failure.file ?? "unknown",
+                name: failure.name ?? "<unnamed test>",
+                title: failure.name ?? "<unnamed test>",
+                suite: failure.suite ?? "",
+                line: failure.line ?? null,
+                error: formatError([failure.error]),
+            });
+        }
+
+        for (const moduleError of report.moduleErrors ?? []) {
+            failingTests.push({
+                file: moduleError.file ?? "unknown",
+                name: "(spec failed to load)",
+                title: "(spec failed to load)",
+                suite: "",
+                line: null,
+                error: formatError([moduleError.error]),
+            });
+        }
+
+        unhandledErrors.push(...(report.unhandledErrors ?? []));
+        testRunSucceeded =
+            failingTests.length === 0 && unhandledErrors.length === 0;
+    } catch (err) {
+        console.warn("Could not parse failure report:", err.message);
+    }
+}
+
+// Fallback: vitest's built-in json report.
+if (!testResultsFound && existsSync(testResultsPath)) {
     try {
         const results = JSON.parse(readFileSync(testResultsPath, "utf8"));
         testResultsFound = true;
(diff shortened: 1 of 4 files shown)


elevatebart:
CI is red here, but not because of this PR — and the comment above is doing its job, so a note on both.

### The red is on `main` too

Run [30527428457](https://github.com/kestra-io/client-sdk/actions/runs/30527428457) on `main` (commit 6d52efc) fails with `8 failed | 418 passed | 13 skipped` across the same 5 spec files. Two independent causes, both pre-existing:

- **`422 … .size: must be less than or equal to 1000`** — `FlowsApi`, `RolesApi`, `ServiceAccountApi` (x2). The specs pass `size: 10000` (e.g. `RolesApi.spec.ts:101`) and the EE image now caps `size` at 1000. `main` was green on [30444520177](https://github.com/kestra-io/client-sdk/actions/runs/30444520177) yesterday with these same specs, so this is the server image moving under us, not a repo change.
- **`Test timed out in 5000ms`** — `set_labels_on_terminated_executions_by_ids` and `searchTestSuitesResultsTest`. Slow e2e tests against the default 5s timeout.

Neither is touched by this diff (a CI comment script, a reporter, and one vitest config flag). Happy to raise a separate PR bumping those four `size` values under the cap and giving the two slow tests an explicit timeout, but it doesn't belong in this one.

### One real bug, fixed

The first comment showed `Error: STACK_TRACE_ERROR` for three of the nine failures. That is vitest's internal sentinel: for a timed-out test the json reporter serialises the failure from the sentinel's stack, so the real `Test timed out in 5000ms.` never lands in `test-results.json` — and the `junit` reporter drops those tests entirely. Timeouts being a common failure mode for these specs, the comment was blank exactly where it mattered.

Fixed in c09543b with a small reporter (`test_javascript_sdk/failureReporter.ts`) that reads the error objects from the reporter API, where the message survives. The comment script prefers its output and falls back to the vitest json report. Verified locally against each shape — timeout, thrown `Error`, assertion failure, nested `describe`, bare-string throw, and a spec that fails to transform — and it reports once per test with the final result, so a flake that passes on retry isn't listed.

The next run should show those two timeouts with their real message instead of the sentinel.

---

jymaire:
Post-merge review notes. The feature works — the comment on this PR rendered correctly and the timeout messages that motivated the custom reporter do come through. Four follow-ups, all in the new code.

### 1. `stripAnsi` doesn't strip the escape byte

`.github/scripts/coverage-comment.mjs:75`

```js
const stripAnsi = (s) => s.replace(/\[[0-9;]*m/g, "");
```

The pattern matches the `[31m` part but not the leading `\x1b`, so a colored error lands in the fenced block with the escape bytes still in it:

```js
stripAnsi("\x1b[31mAssertionError\x1b[39m: expected 1 to equal 2")
// → "\x1bAssertionError\x1b: expected 1 to equal 2"
```

It can also eat legitimate text — `expected [1;2m]` loses the bracket group. Should be:

```js
const stripAnsi = (s) => s.replace(/\x1b\[[0-9;]*m/g, "");
```

Latent on this PR only because the errors that fired happened to be color-free.

### 2. Body truncation can mangle the comment

`coverage-comment.mjs:436`

```js
body = `${body.slice(0, MAX_BODY_CHARS)}\n\n_…comment truncated._…`
```

The cut is at a byte offset with no regard for structure. The comment posted here was 5.5 KB for 12 failures with short errors (~460 chars each); at `MAX_FAILURES_SHOWN = 60` with `MAX_ERROR_CHARS = 600` errors the body lands around 66 KB, so this path will fire on any broad breakage. When it does the cut almost always falls inside a fenced block or an open `<details>`, which swallows the truncation notice and renders the tail as code.

Dropping whole file groups until the body fits (or appending the missing fence / `</details>` closers) keeps it valid markdown.

### 3. Truncation slices before grouping, so the counts disagree with the content

`failingTests` is in completion order, interleaved across files, and `shown` is taken from that:

```js
const shown = failingTests.slice(0, MAX_FAILURES_SHOWN);
```

Then line 361 prints the full per-file count:

```js
<summary>❌ ${fileHeading} — ${tests.length} failing</summary>
```

while `items` above it is filtered down to `shown`. Under truncation you get partial groups from many files, each with a header count larger than the list beneath it. Slicing `failingByFile` instead of `failingTests` gives complete groups and honest counts.

### 4. `failureReporter.ts` is typed `any` where real types exist

`vitest/node` exports `Reporter`, `TestCase` and `TestModule`, and `test_javascript_sdk/tsconfig.json` has `include: ["**/*.ts"]` — so `npm run check:types`, already a step in this workflow, would enforce them:

```ts
import type { Reporter, TestCase, TestModule } from "vitest/node";

export default class FailureReporter implements Reporter {
    onTestCaseResult(testCase: TestCase) { ... }
    onTestRunEnd(testModules: ReadonlyArray<TestModule>, unhandledErrors: ReadonlyArray<unknown>) { ... }
}
```

This file exists specifically to depend on reporter-API internals, so it's the one place where drift on a vitest major should fail at type-check rather than silently produce an empty report. The `typeof module.errors === "function"` guard at line 106 is a symptom of not having the types.

---

For reference, I verified the two load-bearing assumptions in the reporter docblock and both hold: `onTestCaseResult` is dispatched only on the `test-finished` event, which the runner emits once after the retry loop, so a test that passes on attempt 3 is never recorded as a failure; and `failTask` does `result.errors.push(...)` per attempt, so the dedup in `collectErrors` is genuinely needed.

PR #377, opened 2026-07-30
test(javascript): keep page sizes within the API maximum

elevatebart:

The API caps page sizes at 1000 (`PageableUtils.MAX_PAGE_SIZE`) and rejects anything above it with a `422`. Four specs passed `size: 10000` to mean *"every row"* and now fail:

```
422 Invalid entity: searchFlowsBySourceCode.size: must be less than or equal to 1000
422 Invalid entity: searchRoles.size: must be less than or equal to 1000
422 Invalid entity: listServiceAccounts.size: must be less than or equal to 1000
422 Invalid entity: listServiceAccountsForTenant.size: must be less than or equal to 1000
```

- `FlowsApi > search_flows_by_source_code` drops `size` entirely — the search is already narrowed by `q` + `namespace`, so the endpoint's own default is plenty.
- `RolesApi > search_roles` and both `ServiceAccountApi` listings genuinely do want every row, so they ask for `MAX_PAGE_SIZE` — a new constant in `_utils.ts` — rather than an arbitrary large number that drifts out of range the next time the cap changes.

**No SDK behaviour change.** An earlier revision of this PR added a client-side default page size to `configureClient()`; it was dropped. `size` is optional on all 49 paginated endpoints and each carries its own deliberate default — 10 on 41 of them, but 1000 on `listPlugins`, 100 on `listAllMcpServers` / `findDistinctFieldValues`, 1 on the blueprint searches, 25 on the policy-evaluate endpoints. A blanket SDK default would have silently truncated `listPlugins` from 1000 rows to 25. The test edits above are the whole fix.

---

`npm run check:types` is clean after `./generate-sdks.sh javascript`.

CI is the real check here, and `main` is currently red on its own: the run on `6d52efc` ([30527428457](https://github.com/kestra-io/client-sdk/actions/runs/30527428457/job/90821519597)) has 8 failures. This branch removes the four page-size `422`s above. Three failures remain and reproduce on `main` unchanged, unrelated to pagination:

- `ExecutionsApi > pause_executions_by_ids` and `pause_executions_by_query` — `400 invalid bulk pause`, deterministic on both branches, likely a server-side bulk-pause regression deserving its own issue.
- `TestSuitesApi > searchTestSuitesResultsTest` — timeout at 5000ms, also fails on `main`.

### DIFF
--- a/javascript/test_javascript_sdk/FlowsApi.spec.ts
+++ b/javascript/test_javascript_sdk/FlowsApi.spec.ts
@@ -252,7 +252,6 @@ describe('FlowsApi', () => {
         const flow = await createSimpleFlow();
         const resp = await Flows.searchFlowsBySourceCode({
             page: 1,
-            size: 10000,
             q: flow.id,
             namespace: flow.namespace,
         });
--- a/javascript/test_javascript_sdk/RolesApi.spec.ts
+++ b/javascript/test_javascript_sdk/RolesApi.spec.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest';
-import { randomId } from './_utils.js';
+import { MAX_PAGE_SIZE, randomId } from './_utils.js';
 import * as Roles from '@kestra-io/kestra-sdk/roles';
 
 describe('RolesApi', () => {
@@ -98,7 +98,7 @@ describe('RolesApi', () => {
 
         const results = await Roles.searchRoles({
             page: 1,
-            size: 10000,
+            size: MAX_PAGE_SIZE,
             filters: [],
         });
 
--- a/javascript/test_javascript_sdk/ServiceAccountApi.spec.ts
+++ b/javascript/test_javascript_sdk/ServiceAccountApi.spec.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest';
-import { randomIdWith } from './_utils.js';
+import { MAX_PAGE_SIZE, randomIdWith } from './_utils.js';
 import { tenantId } from './_setup.js';
 import * as ServiceAccount from '@kestra-io/kestra-sdk/service-account';
 
@@ -81,7 +81,7 @@ describe('ServiceAccountApi', () => {
         const created = await ServiceAccount.createServiceAccount({ name });
 
         // Many SDKs use {page, size}; some use {page: 1, size: 50}, others 0-based.
-        const results = await ServiceAccount.listServiceAccounts({ page: 1, size: 10000, filters: [] });
+        const results = await ServiceAccount.listServiceAccounts({ page: 1, size: MAX_PAGE_SIZE, filters: [] });
 
         // tolerate different result shapes: {results: []} or direct array
         const items = Array.isArray(results) ? results : results?.results ?? [];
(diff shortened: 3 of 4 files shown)


elevatebart:
### CI status

The first run on this branch ([30527876788](https://github.com/kestra-io/client-sdk/actions/runs/30527876788/job/90822959601)) was red with 3 failures. `main` is red too — the run on `6d52efc` ([30527428457](https://github.com/kestra-io/client-sdk/actions/runs/30527428457/job/90821519597)) has 8. Comparing the two:

| Test | `main` @ 6d52efc | this branch |
|---|---|---|
| `FlowsApi > search_flows_by_source_code` | ❌ `422 … size: must be less than or equal to 1000` | ✅ |
| `RolesApi > search_roles` | ❌ `422 … size: must be less than or equal to 1000` | ✅ |
| `ServiceAccountApi > list_service_accounts` | ❌ `422 … size: must be less than or equal to 1000` | ✅ |
| `ServiceAccountApi > listServiceAccountsForTenant` | ❌ `422 … size: must be less than or equal to 1000` | ✅ |
| `ExecutionsApi > set_labels_on_terminated_executions_by_ids` | ❌ timeout 5000ms | ✅ (flaky) |
| `ExecutionsApi > pause_executions_by_ids` | ❌ `400 invalid bulk pause` | ❌ same |
| `ExecutionsApi > pause_executions_by_query` | ❌ `400 invalid bulk pause` | ❌ same |
| `TestSuitesApi > searchTestSuitesResultsTest` | ❌ timeout 5000ms | ❌ same |

So the four page-size `422`s are the ones this PR fixes, and the three still failing reproduce on `main` unchanged — they predate this branch and are unrelated to pagination:

- **`pause_executions_by_ids` / `pause_executions_by_query` — `400 invalid bulk pause`.** Deterministic (fails all 4 attempts on both branches). Neither endpoint is touched here: `pauseExecutionsByQuery` takes `filters` without `page`/`size`, which this PR deliberately leaves alone, and `pauseExecutionsByIds` sends no query parameters at all. Looks like a bulk-pause regression on the server side rather than an SDK problem — worth its own issue, and I've left it out of this PR rather than papering over it here.
- **`searchTestSuitesResultsTest` — timeout at 5000ms.** Also fails on `main`; a slow/flaky test, not a page-size issue.

I've since rebased onto `6d52efc` (so the branch is tested against the current spec) and pushed a follow-up dropping the `getDefaultPageSize()` accessor — it was only read by the new spec, and every assertion there is already covered by the request URL the SDK emits. A fresh run is underway; the expectation is the same 3 pre-existing failures and nothing else.

---

elevatebart:
Scope reduced — the client-side default page size is dropped, and this PR is now just the test fix. Title and description updated accordingly.

The default was the wrong idea: `size` is already optional on all 49 paginated endpoints, and each carries its own deliberate default rather than one global value.

| server-side `size` default | endpoints |
|---|---|
| 10 | 41 |
| 100 | `listAllMcpServers`, `findDistinctFieldValues` |
| 1000 | `listPlugins` |
| 1 | `searchBlueprints`, `searchInternalBlueprints` |
| 25 | the 3 policy-evaluate endpoints |

A blanket SDK default of 25 would have overridden every one of those — `listPlugins({page: 1})` would have returned 25 plugins instead of 1000, and `findDistinctFieldValues` 25 instead of 100, silently. It also contributed nothing to the failures it was meant to address: those were caused by `size: 10000` exceeding the cap, which only the test edits fix.

---

elevatebart:
moot

PR #378, opened 2026-07-30
test: fix flaky execution pause tests and list pagination

elevatebart:
### What changes are being made and why?

This PR fixes several flaky test issues in the JavaScript SDK test suite:

1. **Execution pause tests** (`ExecutionsApi.spec.ts`):
   - Changed `pause_executions_by_ids` and `pause_executions_by_query` tests to use `LONG_SLEEP_FLOW` instead of `SLEEP_CONCURRENCY_FLOW`
   - The server rejects bulk pause operations with `400 invalid bulk pause` if any execution is no longer RUNNING. Since `SLEEP_CONCURRENCY_FLOW` only sleeps for PT2S (less than the time needed to create all executions), executions were terminating before the bulk pause could complete
   - `LONG_SLEEP_FLOW` (PT30S) ensures executions remain RUNNING throughout the test
   - Added 20-second timeouts to these tests to accommodate the longer sleep duration
   - Added detailed comments explaining the flow selection rationale

2. **List/search pagination tests**:
   - Updated `ServiceAccountApi.spec.ts` list tests to filter by name instead of requesting all results with `size: 10000`
   - Updated `RolesApi.spec.ts` search test to filter by name with `size: 5`
   - Updated `FlowsApi.spec.ts` search test to use `size: 5` instead of `size: 10000`
   - Reason: Server-side page size is capped at 1000, so created test resources can no longer be guaranteed to appear on an unfiltered first page. Filtering by the specific resource name ensures reliable test results.

3. **Test timeouts**:
   - Added 60-second timeout to `TestSuitesApi.spec.ts` test that was timing out
   - Added 20-second timeouts to label-setting tests that now use longer-sleeping executions

### How the changes have been QAed?

The changes are covered by existing unit tests in the SDK test suite. The modifications ensure tests are more resilient by:
- Using appropriate flow sleep durations that match test expectations
- Filtering list/search results to target specific test resources rather than relying on pagination
- Providing adequate timeouts for tests with longer-running operations

All test assertions remain unchanged; only the test setup and timing have been adjusted.

https://claude.ai/code/session_01GwRFP8raLMofiqWs37C6x8

PR #379, opened 2026-07-30
fix: address post-merge review follow-ups on JS SDK test comment (#376)

elevatebart:

Follow-up to #376, addressing [jymaire's post-merge review](https://github.com/kestra-io/client-sdk/pull/376#issuecomment-5129179718). Four issues, all in the code introduced by that PR:

1. **`stripAnsi` embedded a literal ESC control byte in the regex source** (`.github/scripts/coverage-comment.mjs`) instead of the visible `\x1b` escape sequence. It happened to strip ANSI codes correctly at runtime, but the invisible byte is a footgun: GitHub (and some editors) render it as nothing, which is exactly why it read as broken in review. Replaced with the explicit `\x1b\[[0-9;]*m` pattern — same behavior, now visible and safe against tools that don't preserve raw control characters.
2. **Truncation sliced individual tests in completion order** while each file group's header printed the *full* per-file failing count — so under truncation, a group's `— N failing` header could disagree with the entries actually listed beneath it. Truncation now slices whole file groups (at least one group is always shown, even if it alone exceeds the cap), so headers and content always agree.
3. **The final `MAX_BODY_CHARS` cut sliced raw markdown at a byte offset**, which could land inside a fenced code block or an open `<details>`, swallowing the truncation notice and rendering the tail as code. Truncation now drops whole failing-test file groups from the tail until the body fits, falling back to a minimal self-contained notice if it's still too big even with every group dropped.
4. **`failureReporter.ts` typed the vitest reporter-API callbacks as `any`.** `vitest/node` exports `Reporter`, `TestCase`, `TestModule`, and `SerializedError`, so the class now implements `Reporter` with real types. This file exists specifically to depend on reporter-API internals, so it's the one place where drift on a future vitest major should fail at type-check rather than silently produce an empty report — the `typeof module.errors === "function"` guard is gone since `errors()` is no longer optional once properly typed.

- `node --check` on the modified script, plus a standalone `tsc --noEmit` pass against `failureReporter.ts` using the project's compiler options (via `vitest/node`'s shipped `.d.ts`) — no type errors.
- Manually exercised `coverage-comment.mjs` against synthetic failure reports:
  - Normal case (a few failures, no truncation) — output unchanged.
  - Count-truncation case (100 failures across 5 files, `MAX_FAILURES_SHOWN = 60`) — now shows exactly 3 complete file groups (60/60 tests) with an accurate "40 more" note, instead of partial groups with mismatched headers.
  - Size-truncation case (temporarily lowering `MAX_BODY_CHARS` to exercise the path) — confirmed whole groups are dropped from the tail and every `<details>`/code fence in the output stays balanced, with a "report truncated" fallback when even zero groups still don't fit.
  - Verified `stripAnsi` now strips real ANSI sequences (`\x1b[31m...\x1b[39m`) while no longer eating legitimate bracketed text like `[1;2m]`.

No behavior change outside these four fixes; the workflow invocation and vitest reporter registration are untouched.

---
_Generated by [Claude Code](https://claude.ai/code/session_013twU1RFPrfYmvrLk2TZiWv)_

### DIFF
--- a/.github/scripts/coverage-comment.mjs
+++ b/.github/scripts/coverage-comment.mjs
@@ -72,7 +72,7 @@ const sha = COMMIT_SHA || GITHUB_SHA;
 /** Escape the characters that would break a markdown table cell or inline text. */
 const escapeMd = (s) => s.replace(/([|`*_[\]<>])/g, "\\$1");
 
-const stripAnsi = (s) => s.replace(/\[[0-9;]*m/g, "");
+const stripAnsi = (s) => s.replace(/\x1b\[[0-9;]*m/g, "");
 
 /**
  * Permalink to a spec file (optionally a given line) on GitHub. Returns null
@@ -303,7 +303,13 @@ ${items}
 </details>`;
 }
 
-function buildFailingSection() {
+/**
+ * @param dropTrailingGroups How many file groups to additionally drop from
+ *   the tail, beyond the `MAX_FAILURES_SHOWN` cap. Used to shrink the section
+ *   by whole groups when the assembled body is still over the size limit,
+ *   instead of slicing the final markdown at an arbitrary byte offset.
+ */
+function buildFailingSection(dropTrailingGroups = 0) {
     if (failingTests.length === 0 && unhandledErrors.length > 0) {
         return `### ❌ JavaScript SDK — tests failed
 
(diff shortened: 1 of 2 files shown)


elevatebart:
CI is red here, but not from this diff — it's the same pre-existing flakiness already tracked in #378 (open): `400 invalid bulk pause` (execution no longer RUNNING before the bulk op completes), `422 …size: must be less than or equal to 1000` (EE page-size cap), and the e2e timeouts. Every failing test in this run's report is fixed by #378's changes to `ExecutionsApi.spec.ts`, `FlowsApi.spec.ts`, `RolesApi.spec.ts`, `ServiceAccountApi.spec.ts`, and `TestSuitesApi.spec.ts`.

This PR only touches `.github/scripts/coverage-comment.mjs` and `failureReporter.ts` — no test specs or SDK client code — so it can't affect these results either way. Will re-check once #378 merges and this branch is rebased onto a green base.

---

jymaire (approved):

PR #393, opened 2026-08-27
fix(go): add /v2 module suffix required for the v2.0.0-rc1 tag

fdelbrayelle:
## Summary

The `go-sdk/v2.0.0-rc1` tag push failed release CI ([run 33079040023](https://github.com/kestra-io/client-sdk/actions/runs/33079040023), exit 22): `proxy.golang.org` rejects it because `go.mod`'s module path lacked the `/v2` suffix Go requires for major version >= 2.

> not found: ...go-sdk@v2.0.0-rc1: invalid version: module contains a go.mod file, so module path must match major version ("github.com/kestra-io/client-sdk/go-sdk/v2")

This bumps the module path to `github.com/kestra-io/client-sdk/go-sdk/v2` and updates every internal import and doc reference, then makes the release workflow major-aware.

## Strategy: major branch, not major subdirectory

Both are legal per the [module reference](https://go.dev/ref/mod) — "the module subdirectory is the part of the module path that names the directory, **not including the major version suffix**", so `.../go-sdk/v2` may live in either `go-sdk/` or `go-sdk/v2/`.

Keeping it in `go-sdk/` means the tag stays `go-sdk/vX.Y.Z` for both majors. Duplicating into `go-sdk/v2/` would force `go-sdk/v2/v2.0.0` and break the `${GITHUB_REF_NAME#go-sdk/}` strip.

- `main` -> v2 code, module path `.../go-sdk/v2`, tagged `go-sdk/v2.0.0`
- `releases/v1.3.x` -> v1 code, module path unchanged, tagged `go-sdk/v1.3.x`

`releases/v1.3.x` already carries the bare module path and does not contain 138f333c (the breaking Kestra 2.0 commit), so it is ready to cut v1 patches from as-is.

## Release workflow

The indexing URLs now derive the major suffix from the tag rather than hardcoding it, so v1 tags cut from the maintenance branch still resolve (v0/v1 bare, v2+ append `/vN`). The `pkg.go.dev` warm-up is now best effort — an unindexed brand-new version previously failed the job and skipped the Slack announcement.

> [!WARNING]
> Once this merges, a v1.x tag must be cut from `releases/v1.3.x` or an older commit — never from `main`. Tagging `go-sdk/v1.3.1` at a revision whose `go.mod` says `/v2` gets the mirror image of the error above.

## Test plan

- [x] `go build ./...` and `go vet ./...` clean in `go-sdk/`
- [x] Release workflow YAML parses; tag parsing verified for `v1.3.1`, `v2.0.0-rc1`, `v2.0.0`, `v0.9.0`, `v10.1.0` (correct path each, no early exit under `bash -e`)
- [x] `EmbedSnippets --check README.md` exit 0 (the injected Go block carries no import line, so it is unaffected)
- [ ] Re-push a `go-sdk/v2.0.0-rc1` tag after merge and confirm `Announce Go SDK release` succeeds

### DIFF
--- a/.github/workflows/go-sdk-release.yml
+++ b/.github/workflows/go-sdk-release.yml
@@ -16,8 +16,8 @@ jobs:
       - name: Trigger pkg.go.dev indexing
         run: |
           VERSION="${GITHUB_REF_NAME#go-sdk/}"
-          curl -sf "https://proxy.golang.org/github.com/kestra-io/client-sdk/go-sdk/@v/${VERSION}.info" > /dev/null
-          curl -sf "https://pkg.go.dev/github.com/kestra-io/client-sdk/go-sdk@${VERSION}" > /dev/null
+          curl -sf "https://proxy.golang.org/github.com/kestra-io/client-sdk/go-sdk/v2/@v/${VERSION}.info" > /dev/null
+          curl -sf "https://pkg.go.dev/github.com/kestra-io/client-sdk/go-sdk/v2@${VERSION}" > /dev/null
 
       - name: Slack notification
         uses: kestra-io/actions/composite/slack-status@main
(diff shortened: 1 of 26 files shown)


AcevedoR (approved):
finger crossed, hoping it will not be necessary too to move the project into a v2 path AND change the artifact name

PR #396, opened 2026-08-27
refactor(javascript): move the SDK barrel off the root entry onto /all

elevatebart:

Puts this SDK on the same entry-point shape as the in-repo OSS/EE UI SDKs, which had to make this split for a build reason: their root entry is a module-federation share, so MF pins every name it exports and re-exporting the operations there held all per-tag modules in the host's initial graph (~228 KB). That constraint doesn't apply here — this package is plain npm and consumers tree-shake normally — so this is for consistency: the same import should work whichever SDK you're holding.

- Root entry: client facade (`useClient` / `configureClient` / `setMockClient`) + generated **types** only.
- New `@kestra-io/kestra-sdk/all`: every operation — named exports, plus the namespace as `default`.
- `README_JAVASCRIPT_SDK.md` gains an entry-point table.

**BREAKING CHANGE:** no operations on the package root. Use its module subpath (`@kestra-io/kestra-sdk/flows`) — the only form this guide documents — or `/all` for a one-line migration. Types, `/client` and per-module entries unaffected. This is the published SDK, so it wants a major version.

Also carries `08f7e0e`: the JavaScript half of #397, ported to unblock CI here. Kestra 2.0 stopped auto-granting tenant access on group add, so 5 tests failed with `404 User does not exist` — unrelated to this PR, and a no-op once #397 lands.

---

Regenerated from the committed spec, then `npm run build`, `npm run check:types` and CI's own suite — all green. Loading the built `dist/`: root exposes exactly the 3 client functions and no operation; `/all` exposes 1188 named exports and its `default` is the namespace. `tsdown`'s `exports: true` regenerated `package.json` with `"./all"`; that's included.

Unrelated: `npm run typecheck` in `javascript-sdk` is broken on `main` — it calls `vue-tsc`, which the package doesn't install. Worth a separate fix.

---

### Setup Instructions

None.

## Related PRs

- [kestra-io/kestra#18754](https://github.com/kestra-io/kestra/pull/18754) — OSS UI SDK split, plus two unrelated warning fixes
- [kestra-io/kestra-ee#10528](https://github.com/kestra-io/kestra-ee/pull/10528) — EE UI SDK split and the federation share config

https://claude.ai/code/session_01UvrhcR8x2kPEDVaf1E6XJS

### DIFF
--- a/README_JAVASCRIPT_SDK.md
+++ b/README_JAVASCRIPT_SDK.md
@@ -15,6 +15,18 @@ npm install @kestra-io/kestra-sdk
 
 The SDK is organized into domain-specific modules (e.g. `flows`, `executions`, `apps`, `plugins`). Import only the modules you need to keep your bundle lean and benefit from tree-shaking.
 
+### Entry points
+
+| Import | Contents |
+| --- | --- |
+| `@kestra-io/kestra-sdk` | `useClient` / `configureClient` / `setMockClient`, and every generated **type** |
+| `@kestra-io/kestra-sdk/client` | the shared HTTP client |
+| `@kestra-io/kestra-sdk/<module>` | the operations of one module, e.g. `/flows`, `/executions` |
+| `@kestra-io/kestra-sdk/all` | every operation at once — named exports, plus the namespace as `default` |
+
+The root entry exports no operations. Reach an operation through its module (the tree-shakeable
+form used throughout this guide), or through `/all` if one import is preferable to several.
+
 ### Configure the client
 
 Import and configure the shared HTTP client once, at the entry point of your application:
--- a/javascript/javascript-sdk/package.json
+++ b/javascript/javascript-sdk/package.json
@@ -12,6 +12,7 @@
     ".": "./dist/index.js",
     "./ai": "./dist/ai.js",
     "./ai-admin": "./dist/ai-admin.js",
+    "./all": "./dist/all.js",
     "./apps": "./dist/apps.js",
     "./assets": "./dist/assets.js",
     "./audit-logs": "./dist/audit-logs.js",
--- a/javascript/javascript-sdk/src/all.ts
+++ b/javascript/javascript-sdk/src/all.ts
@@ -1,0 +1,8 @@
+// The whole generated API surface behind one specifier, for consumers who prefer a single import
+// over the per-tag subpaths. This is the only place the barrel lives: the root entry exports the
+// client facade and the generated types, keeping the three Kestra SDKs (this one and the in-repo
+// OSS/EE UI SDKs) on the same entry-point shape.
+import * as sdk from "./openapi/index"
+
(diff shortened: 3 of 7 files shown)


elevatebart:
`Test Javascript SDK` failed on `1f2fc85` — 5 tests, all `404 Not Found: User does not exist for id`, in `GroupsApi.spec.ts` (4) and `UsersApi.spec.ts` (1).

Not this PR's. It's the Kestra 2.0 server change that stopped auto-granting tenant access when a user is added to a group, so the tenant-scoped group endpoints reject the user. Two things rule this branch out: the failing specs import their operations from module subpaths (`@kestra-io/kestra-sdk/groups`, `/users`) and take only *types* from the root, both of which this PR preserves; and the failures are live-server 404s, not module-resolution errors.

[#397](https://github.com/kestra-io/client-sdk/pull/397) fixes it — same root cause as the Go/Java fix in 0d7a127, and its description notes neither the Python nor JavaScript suite had run since the server change, so both would fail on their next execution. I've ported its **JavaScript half** into `08f7e0e` rather than wait: `tenants: [tenantId]` at user creation, byte-identical to #397's JS diff. It no-ops once #397 lands. I deliberately left #397's Python half out, since the failing check here doesn't need it.

Verified before pushing: the ported diff matches #397's JS changes exactly, and `tsc --noEmit` on the test workspace reports 46 errors both with and without the port, none mentioning `tenants` — so the field is valid on `createUser` and the port introduces nothing. (Those 46 are pre-existing `string | undefined` strictness errors across the specs, unrelated to either change; the workspace has no typecheck in CI.)

I couldn't run the suite itself here — it needs a live Kestra instance — so CI on `08f7e0e` is the real signal. I'm watching it.

---

elevatebart:
Correcting one thing in my comment above: I said `tsc --noEmit` on the test workspace reports 46 errors with and without the ported fix. That was my own bad invocation (I pointed `tsc` at the tsconfig by path from the wrong directory). Running what CI actually runs — `npm run check:types` in `javascript/`, i.e. `npm run typecheck --workspace @kestra-io/kestra-sdk-test` — the workspace type-checks **clean, exit 0**, on this branch including the port.

There are no pre-existing type errors in the test workspace; please disregard that number. Everything else in that comment stands: the tenant-access 404s are the Kestra 2.0 server change, not this PR, and `08f7e0e` carries the JavaScript half of #397 verbatim.

Worth noting for #389 (`chore/update-openapi-spec`) while I'm here: its failures are at this same `type check` step, before the tests run — consistent with the spec update renaming operations (`searchConcurrencyLimits` → `searchConcurrencyLimitsasInstanceOwner`, `updateConcurrencyLimit` → `updateConcurrencyLimitasInstanceOwner`) while the committed specs still reference the old names. Separate from this PR, but it's why `kestra-ee.yml` here is behind EE `develop`.

---

PR #403, opened 2026-08-31
feat: close the last Go/Java coverage gaps against the rc12 spec

AcevedoR:
Auditing both hand-written SDKs against `kestra-ee.yml` — which is current with `v2.0.0-rc12`, no route drift — left six routes bound in one language but not the other. This closes all six, three per side.

### Go gains what Java already had

| Route | Method |
|---|---|
| `POST /{tenant}/cases/from-task` | `CasesAPI.CreateCaseFromTask` |
| `POST /{tenant}/executions/{id}/actions/replay-with-inputs` | `ExecutionsAPI.ReplayExecutionWithInputs` |
| `GET /{tenant}/apps/view/{id}/streams/{stream}` | `AppsAPI.StreamAppEvents` |

Cases had no Go binding at all, so `CasesAPI` is new and wired onto `KestraClient`. `from-task` takes an `interface{}` body because no Go model is generated for `CasesController.CaseFromTaskRequest`. `ReplayExecutionWithInputs` sends inputs as multipart form fields, which is what the endpoint expects; `ReplayExecution` remains the call for reusing the original inputs.

### Java gains what Go already had

| Route | Method |
|---|---|
| `POST /{tenant}/blueprints/flows/validate` | `BlueprintsApi.validateFlowBlueprint` |
| `POST /{tenant}/bindings/bulk` | `BindingsApi.bulkCreateBindings` |
| `GET /api/v1/configs` | `MiscApi.configuration` |

`/api/v1/configs` is instance-level and had no home, so `MiscApi` is new and wired onto `KestraClient`.

### Two tests that assert less than you might expect

Both are deliberate, and both still fail if the binding is wrong:

- **`streamAppEvents`** asserts the channel *closes* rather than that the call errors. The server answers a stream subscription with `200` and an SSE body even for an unknown app, so a bad path — not a missing app — is what surfaces as an error.
- **`validateFlowBlueprint`** only asserts a valid source is accepted. That endpoint returns `constraints=null` for every input on the `-no-plugins` image, including malformed YAML and unknown task types, so a negative assertion would pin nothing.

### Verification

Against `kestra-ee:v2.0.0-rc12`:

- Full `go-sdk` suite green.
- Full `java-sdk-test` suite green — **430 tests, 0 failures, 15 skipped.**
- Each of the six new bindings was neutered in turn (path pointed at a wrong segment) and the covering test failed in every case, so none of them is vacuous.

After this, both SDKs bind the same set of routes and neither has a route the rc12 spec no longer serves.

### DIFF
--- a/go-sdk/kestra_api_client/apps_api.go
+++ b/go-sdk/kestra_api_client/apps_api.go
@@ -89,3 +89,9 @@ func (a *AppsAPI) LogsFromAppExecution(ctx context.Context, uid, tenant string, 
 	appendRepeatedParam(params, "taskIds", taskIds)
 	return a.doDownload(ctx, "GET", tenantPath(tenant, "apps", "view", uid, "logs", "download"), nil, params)
 }
+
+// StreamAppEventsFromApp follows an app's event stream. The channel is closed when
+// the stream ends or ctx is cancelled.
+func (a *AppsAPI) StreamAppEvents(ctx context.Context, id, stream, tenant string) (<-chan *EventAppResponse, error) {
+	return followSSE[EventAppResponse](&a.baseAPI, ctx, tenantPath(tenant, "apps", "view", id, "streams", stream), nil)
+}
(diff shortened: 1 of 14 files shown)


brian-mulier-p (approved):
LGTM

PR #408, opened 2026-09-02
fix(java): allow building an unauthenticated KestraClient

### LINKED ISSUE #407: KestraClientBuilder cannot build an unauthenticated client
## Bug

`KestraClient.KestraClientBuilder` has no way to produce a client that sends no `Authorization` header.

The builder's constructor initialises `auth` to `Auth.BASIC`, and `build()` switches on `auth.ordinal()`. With neither `tokenAuth(...)` nor `basicAuth(...)` called, the BASIC branch base64-encodes `username + ":" + password` with both still `null`, so every request carries `Authorization: Basic bnVsbDpudWxs` — base64 of the literal string `null:null`.

`Auth.NONE` exists and `build()` already treats it as the no-header case, but nothing can select it: the builder only exposes `url`, `tokenAuth`, `basicAuth`, `connectTimeout`, `readTimeout` and `build`, and the `KestraClient(ApiClient)` constructor is package-private.

### Steps to Reproduce

1. `KestraClient.builder().url("http://localhost:8080").build()`
2. Send any request with the resulting client
3. Inspect the request headers

### Expected Behavior

A builder without credentials produces a client that sends no `Authorization` header, so an unsecured Kestra API can be called. A `noAuth()` (or a public `auth(Auth)`) setter would express this.

### Actual Behavior

Every request carries `Authorization: Basic bnVsbDpudWxs`. Against an unsecured API this is a malformed credential that happens to be ignored; against a secured one the caller gets an opaque 401 instead of a clear configuration error.

Verified in the published bytecode of `1.3.2`, `2.0.0-rc2` and `2.0.0-rc3`.

### Impact

Plugin tasks cannot offer an "unsecured API" option. `kestra-io/plugin-ai#414`, `kestra-io/plugin-kestra#181`, `kestra-io/plugin-git#331` and `kestra-io/plugin-ee-git#164` each tried to and had to revert it.

brian-mulier-p:
Closes https://github.com/kestra-io/client-sdk/issues/407.

`KestraClientBuilder`'s constructor initialised `auth = Auth.BASIC`. With neither
`basicAuth(...)` nor `tokenAuth(...)` called, `build()` took the BASIC branch with both
credentials still `null` and emitted `Authorization: Basic bnVsbDpudWxs` — base64 of the
literal string `null:null`. `Auth.NONE` existed and was already the no-header case, but
nothing could select it.

### Changes

- `KestraClientBuilder.noAuth()` selects `Auth.NONE`, and can reset a previously set auth.
- The builder default becomes `Auth.NONE`.
- `ApiClient.getDefaultHeaders()` returns an unmodifiable view of the default headers, so the
  test can assert on what the client actually sends.
- README: document the no-auth path, and fix the timeouts snippet that used a non-existent
  `.token(...)` builder method instead of `.tokenAuth(...)`.

`noAuth()` rather than a public `auth(Auth)`: it mirrors the existing `tokenAuth`/`basicAuth`
mode-named setters that assign `this.auth` themselves, and a public `auth(Auth.TOKEN)` would
re-open the same hole — selecting a mode with no credentials behind it.

`ApiClient.java` is in the diff only for that getter. It is hand-edited already (#275 added the
timeout setters) and `generate-sdks.sh` refuses `java`, so there is no generator to overwrite it.

### ⚠️ Signature change

No public method signature changed. One default behavior did:

`KestraClientBuilder` default auth `Auth.BASIC` → `Auth.NONE`, so
`KestraClient.builder().url(x).build()` now sends no `Authorization` header instead of
`Authorization: Basic bnVsbDpudWxs`.

Not considered a break: against a secured instance both the old and the new behavior yield a
401; against an unsecured one the old header was simply ignored. Every
`KestraClient.builder()` usage in this repo (tests, `docs/*.md` samples) sets auth explicitly,
so none depended on the default. The flip also makes the bug unconstructable rather than merely
unlikely — `Auth.BASIC` is now reachable only through `basicAuth(...)`, which `requireNonNull`s
both arguments.

### Verification

The new test asserts on the built client's actual default headers, not on `notNull`. Before the
fix it failed with:

    Expecting actual:
      {"Authorization"="Basic bnVsbDpudWxs", "User-Agent"="OpenAPI-Generator/v1.0.5/java"}
    not to contain key:
      "Authorization"

`KestraClientAuthTest` covers: no credentials, explicit `noAuth()`, `noAuth()` after
`basicAuth(...)`, the exact base64 for basic, and the exact `Bearer` value for token.

Ran `compileJava` + `compileTestJava` over `java-sdk-test`, then `KestraClientAuthTest` (5) and
`KestraClientTimeoutTest` (4) — all green. The remaining `java-sdk-test` classes need the docker
Kestra instance from `run-tests.sh` and were not run.

Once published, this unblocks the "unsecured Kestra API" option that kestra-io/plugin-ai#414,
kestra-io/plugin-kestra#181, kestra-io/plugin-git#331 and kestra-io/plugin-ee-git#164 each had to
revert.

### DIFF
--- a/java/java-sdk/src/main/java/io/kestra/sdk/internal/ApiClient.java
+++ b/java/java-sdk/src/main/java/io/kestra/sdk/internal/ApiClient.java
@@ -425,6 +425,15 @@ public class ApiClient extends JavaTimeFormatter {
   }
 
   /**
+   * Get the default headers sent with every request.
+   *
+   * @return An immutable snapshot of the default headers
+   */
+  public Map<String, String> getDefaultHeaders() {
+    return Map.copyOf(defaultHeaderMap);
+  }
+
+  /**
    * Add a default cookie.
    *
    * @param key The cookie's key
@@ -508,7 +517,7 @@ public class ApiClient extends JavaTimeFormatter {
 
   /**
    * Set the date format used to parse/format date parameters.
-   * @param dateFormat Date format
+   * @param newDateFormat Date format
    * @return API client
    */
   public ApiClient setDateFormat(DateFormat newDateFormat) {
--- a/java/java-sdk/README.md
+++ b/java/java-sdk/README.md
@@ -853,6 +853,11 @@ Authentication schemes defined for the API:
 
 - **Type**: HTTP Bearer Token authentication (Bearer)
 
+Both are optional: the `KestraClient` builder sends no `Authorization` header unless
+`basicAuth(username, password)` or `tokenAuth(token)` is called, so an unsecured Kestra API can
+be reached with `KestraClient.builder().url("http://localhost:8080").build()`. `noAuth()` states
+this explicitly and resets any previously configured authentication.
+
 
 ## Configuring timeouts
 
@@ -865,7 +870,7 @@ import java.time.Duration;
 
 var client = KestraClient.builder()
     .url("https://kestra.example.com")
-    .token("your-api-token")
+    .tokenAuth("your-api-token")
     .connectTimeout(Duration.ofSeconds(10))   // 0 = infinite (default)
     .readTimeout(Duration.ofMinutes(30))       // 0 = infinite (default)
     .build();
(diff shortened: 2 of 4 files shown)


java/java-sdk/src/main/java/io/kestra/sdk/internal/ApiClient.java
AcevedoR: minor jdoc fix

AcevedoR (approved):

PR #259, opened 2026-06-02
feat(go): non-breaking dual client (restore v1 surface alongside hand-written)

jymaire:
## Why
The hand-written client rewrite (#230 + follow-ups) was API-breaking vs the generated `go-sdk/v1.0.x`, which would have forced a `v2` major bump and a full migration for every consumer (kestra-io/kestractl#62). This PR instead makes the next release **non-breaking**: it exposes the generated v1 surface *and* the new hand-written API in one package on the `v1.x` module path.

Supersedes #257 (the v2 module-path bump).

## What it does
- **Restores the generated v1 surface** from `go-sdk/v1.0.0` — transport (`legacy_client.go`/`legacy_configuration.go`/`legacy_response.go`) + the 12 `api_*.go` services + the data models the rewrite had pruned. So `NewConfiguration`/`NewAPIClient`/`client.API.GroupsAPI.X().Execute()`/`GenericOpenAPIError`/`ContextAccessToken`/`BasicAuth` all exist again.
- **Keeps the new hand-written API** (`NewClient` / `client.Groups()` / `*ApiError` / `WithDebug` / invitations / bindings) — dual API, additive.
- **Resolves shared-type conflicts** so v1 consumers compile unchanged:
  - Renamed the hand-written `QueryFilter`→`SearchFilter` (+`SearchFilterField`/`SearchFilterOp`); restored the generated `QueryFilter` models.
  - Restored v1.0.0 shape of `KVControllerKvDetail.Value` (`map[string]interface{}`) and `ApiToken` (optional fields).
  - Added a type alias `IAMRoleControllerApiRoleCreateOrUpdateRequestPermissions = ListPermissions200Response` (`compat_aliases.go`).
- Added `go-sse v0.11.0` (generated executions SSE dependency).

## Verification
- **Drop-in proof:** with a temporary `replace`, **kestractl builds and all 218 unit tests pass with zero kestractl code changes.**
- SDK: `go build ./...` ✅, `go test ./kestra_api_client/` ✅, `go test -c ./test/` ✅. `go vet` clean except pre-existing `api_executions.go` context-cancel warnings (shipped in v1.0.0).
- Live-EE integration tests (`./test/`) need a running Kestra and will run in CI.

## Release impact
Ship as a **non-breaking minor (`v1.1.0`)** — no `/v2`, no consumer migration. Existing `v1.0.x` code keeps working; new code can adopt the hand-written API.

## Notes / follow-ups
[...]

PR #409, opened 2026-09-03
fix(go): route the single-execution actions to their 2.0 /actions/ paths

AcevedoR:
## What

`api_executions.go` was still on the Kestra 1.x URL scheme. Two fixes, both bringing it back in line with `kestra-ee.yml`:

1. The 11 single-execution actions move under `/executions/{executionId}/actions/`.
2. The 10 executions action bulk methods are retyped from `*BulkResponse` to `*ApiAsyncOperationResponse`.

## Why it drifted

`api_executions.go` was restored from the pre-2.0 generated output in #259, to keep the v1 request-builder surface alive alongside the new hand-written client. The 2.0 route move never reached it. Since the Go generator config was removed in #230, nothing regenerated it either.

The spec has been right the whole time — so the generated Python/JavaScript SDKs are unaffected. This is isolated to the Go v1-compat surface.

## The symptom is misleading

On 2.0 only `kill` returns something recognisable (403). The POST actions fall through to `POST /executions/{namespace}/{flowId}`, so:

```
POST /api/v1/main/executions/{id}/replay          -> 404 "Requested Flow is not found."
POST /api/v1/main/executions/{id}/actions/replay  -> 200
```

That 404 is the server looking for a *flow* named `replay`. It reads like a missing execution, which is what sent the original report looking in the wrong place.

The mirror image already existed: the hand-written `executions_api.go` was written against the 2.0 spec, so its `eval` 403s on 1.3. **Neither half of the SDK worked on both servers** — the generated half only on 1.3, the hand-written half only on 2.0.

## Scope

| | |
|---|---|
| **Re-pathed (11)** | `change-status`, `force-run`, `kill`, `labels`, `pause`, `replay`, `replay-with-inputs`, `restart`, `resume`, `state`, `unqueue` |
| **Untouched — did not move in 2.0** | `file`, `file/metas`, `flow`, `follow`, `follow-dependencies`, `graph` |
| **Untouched — bulk paths never moved** | `<action>/by-ids`, `<action>/by-query` |

All confirmed against a live 2.0.0-rc13, not inferred from the spec. After this change every path in `api_executions.go` resolves in `kestra-ee.yml`.

## The bulk retype

The executions action bulk endpoints answer with the 2.0 async-operation shape; the spec types all 18 as `ApiAsyncOperationResponse`. `BulkResponse` only has `count`, so the number was dropped at decode — the call succeeded, the executions were killed, and the caller read **0**. A silent wrong answer rather than an error.

```
DELETE /api/v1/main/executions/kill/by-ids
  2.0.0-rc13 -> 202 {"operationId":"2gtoTYnQy1KHopwSQAXE3y","totalItems":1}
  1.3.35     -> 200 {"count":1}
```

`DeleteExecutionsByIds` stays `*BulkResponse` — the spec still types the *delete* bulk endpoints that way, so `count` is correct there. The remaining `by-query` variants already return `map[string]interface{}`; they're untyped rather than mistyped, so I left them alone.

## ⚠️ Signature change

Per AGENTS.md. Return type only — parameters and arity are unchanged on every method.

`(*BulkResponse, *http.Response, error)` → `(*ApiAsyncOperationResponse, *http.Response, error)`

on `ForceRunByIds`, `KillExecutionsByIds`, `PauseExecutionsByIds`, `ReplayExecutionsByIds`, `RestartExecutionsByIds`, `ResumeExecutionsByIds`, `SetLabelsOnTerminatedExecutionsByIds`, `UnqueueExecutionsByIds`, `UpdateExecutionsStatusByIds`, `UpdateExecutionsStatusByQuery` (and their `*Execute` counterparts).

Callers read `GetTotalItems()` instead of `GetCount()`. This is a compile-time break, which is the intent — the value they were reading was always 0 on 2.0. **v2 is still rc, so this lands before GA rather than after.** I measured the blast radius on the one known consumer, kestractl: 9 call sites, all loud compile errors, no silent behaviour change. Happy to split this commit out if you'd rather take the path fix alone first.

## Verification

I could not run `go-sdk/run-tests.sh` locally — it needs the EE licence CI secret for `test-utils/docker-setup/application-secrets.yml`, so **CI needs to confirm the suite**. What I did run:

- `go build ./...` and `go vet ./kestra_api_client/` clean.
- End-to-end against **live EE 2.0.0-rc13 and EE 1.3.35 side by side**, driving every affected method through kestractl built against this branch: `kill`, `pause`, `force-run`, `replay`, `restart`, `resume`, `eval-expression`, `replay-with-inputs`, `set-labels`, `change-status`, plus the `-bulk` and `-by-query` families. All pass on 2.0, with correct bulk counts.
- Each action was exercised against a correctly-stated execution (a FAILED one for `restart`, a genuinely paused one for `resume`), so a state conflict could not be mistaken for a passing route.

One consequence worth flagging for downstream: with this branch, a 2.x-targeting SDK reading `totalItems` gets 0 from a 1.3 server, which still sends `count`. That's expected — this SDK targets 2.x — and consumers wanting one binary against both servers need a compat shim either way. kestractl has one.

Refs kestra-io/kestractl#117

### DIFF
--- a/go-sdk/kestra_api_client/api_executions.go
+++ b/go-sdk/kestra_api_client/api_executions.go
@@ -1994,7 +1994,7 @@ func (a *ExecutionsAPIService) ForceRunExecutionExecute(r ApiForceRunExecutionRe
 		return localVarReturnValue, nil, &GenericOpenAPIError{error: err.Error()}
 	}
 
-	localVarPath := localBasePath + "/api/v1/{tenant}/executions/{executionId}/force-run"
+	localVarPath := localBasePath + "/api/v1/{tenant}/executions/{executionId}/actions/force-run"
 	localVarPath = strings.Replace(localVarPath, "{"+"executionId"+"}", url.PathEscape(parameterValueToString(r.executionId, "executionId")), -1)
 	localVarPath = strings.Replace(localVarPath, "{"+"tenant"+"}", url.PathEscape(parameterValueToString(r.tenant, "tenant")), -1)
 
@@ -2237,7 +2237,7 @@ func (a *ExecutionsAPIService) KillExecutionExecute(r ApiKillExecutionRequest) (
 		return localVarReturnValue, nil, &GenericOpenAPIError{error: err.Error()}
 	}
 
-	localVarPath := localBasePath + "/api/v1/{tenant}/executions/{executionId}/kill"
+	localVarPath := localBasePath + "/api/v1/{tenant}/executions/{executionId}/actions/kill"
 	localVarPath = strings.Replace(localVarPath, "{"+"executionId"+"}", url.PathEscape(parameterValueToString(r.executionId, "executionId")), -1)
 	localVarPath = strings.Replace(localVarPath, "{"+"tenant"+"}", url.PathEscape(parameterValueToString(r.tenant, "tenant")), -1)
 
@@ -2721,7 +2721,7 @@ func (a *ExecutionsAPIService) PauseExecutionExecute(r ApiPauseExecutionRequest)
 		return nil, &GenericOpenAPIError{error: err.Error()}
 	}
 
-	localVarPath := localBasePath + "/api/v1/{tenant}/executions/{executionId}/pause"
+	localVarPath := localBasePath + "/api/v1/{tenant}/executions/{executionId}/actions/pause"
 	localVarPath = strings.Replace(localVarPath, "{"+"executionId"+"}", url.PathEscape(parameterValueToString(r.executionId, "executionId")), -1)
 	localVarPath = strings.Replace(localVarPath, "{"+"tenant"+"}", url.PathEscape(parameterValueToString(r.tenant, "tenant")), -1)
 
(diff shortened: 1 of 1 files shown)


AcevedoR:
Some corroboration I dug up after opening this, which I think settles the retype — it isn't a judgement call on my part, it's already this package's own convention.

**The hand-written client in the same file draws exactly the same line:**

| Method | `executions_api.go` (hand-written, 2.0-era) | `api_executions.go` before → after |
|---|---|---|
| `KillExecutionsByIds` | `*ApiAsyncOperationResponse` | `*BulkResponse` → `*ApiAsyncOperationResponse` |
| `PauseExecutionsByIds` | `*ApiAsyncOperationResponse` | `*BulkResponse` → `*ApiAsyncOperationResponse` |
| `ForceRunByIds` | `*ApiAsyncOperationResponse` | `*BulkResponse` → `*ApiAsyncOperationResponse` |
| `UpdateExecutionsStatusByIds` | `*ApiAsyncOperationResponse` | `*BulkResponse` → `*ApiAsyncOperationResponse` |
| `DeleteExecutionsByIds` | `*BulkResponse` | `*BulkResponse` (unchanged) |

The action/delete split I drew is the split the hand-written client already had — including keeping `DeleteExecutionsByIds` on `BulkResponse`. This PR only makes the generated surface agree with the hand-written one.

**And the SDK's own tests already assert that shape**, via the hand-written client:

```go
// test/api_executions_test.go:482
res, err := KestraTestClient().Executions().KillExecutionsByIds(ctx, MAIN_TENANT, []string{exec1.Id, exec2.Id})
require.EqualValues(t, 2, res.GetTotalItems())
```

Ten such `GetTotalItems()` assertions across `api_executions_test.go`. So `totalItems` is not something I inferred from the spec — it's what the suite has been asserting on a live server all along. The generated surface was simply the one place still on `count`.

**No SDK test breaks:** `go test -c ./test/` compiles clean, and no test touches the generated `ExecutionsAPIService` bulk methods, so nothing in the suite reads `GetCount()` on a retyped method.

**Mechanical checks against the spec**, in case useful for review:

- All **45** path+verb pairs in `api_executions.go` now resolve in `kestra-ee.yml` (before: 11 paths did not exist in the spec at all).
- All **20** executions bulk methods' return types match the spec's response schema — 0 mismatches.

Both were scripted cross-references against the spec rather than eyeballed.

PR #412, opened 2026-09-04
fix(go)!: bind /acls/permissions and stop decoding actions into a closed enum

AcevedoR:
## What was wrong

`MiscAPI.ListActions` calls `GET /acls/actions`, which **Kestra 2.0 removed** when `#7550` replaced the flat action vocabulary with the per-resource `Resource`/`Action` model. Against 2.0 the call 403s, so kestractl's `server actions` is dead there — the same 403 route-miss signature as kestra-io/kestractl#117.

The replacement is `GET /acls/permissions` (`listPermissions` in `kestra-ee.yml`). This PR binds it and keeps both methods.

## Why both, and why no fallback is possible

This is **not** a path rename you can shim. `/acls/permissions` exists on *both* server lines and returns incompatible bodies. From `AclController` on `releases/v1.3.x` vs `develop`:

| | 1.3 | 2.0 |
|---|---|---|
| `GET /acls/actions` | `List<Action>` | **removed** |
| `GET /acls/permissions` | `List<Permission>` | `Map<Resource, Set<Action>>` |

Verified on the wire:

```
1.3.35      GET /acls/actions      -> ["READ","CREATE","UPDATE","DELETE"]
            GET /acls/permissions  -> ["FLOW","BLUEPRINT","TEMPLATE", ...]
2.0.0-rc13  GET /acls/actions      -> 403
            GET /acls/permissions  -> {"FLOW":["VIEW",...],"ASSET":[...], ...}
```

Same path, different *kind* of answer. A transport-level rewrite of `actions`→`permissions` would, on 1.x, silently hand the caller a list of resource **names** rendered as actions — wrong data, no error. So choosing by server version is the only correct option, and both doc comments say so explicitly.

## Actions are now strings, not the `Action` enum

These endpoints exist to report *the server's own* action vocabulary. Decoding them into a generated closed enum defeats that: any value the enum doesn't know fails the entire call. That isn't hypothetical in either direction:

- **1.3** returns `READ`, which `Action` has never contained — so *every* 1.3 call failed to decode despite a 200. kestractl already carried a workaround, recovering the raw list out of the error body. It can now delete it.
- **2.0** returns `LOCK`, `PROMOTE` and `TEMPLATE`, none of which `Action` contains.

Adding those members would fix today and break on the next server that adds an action. Strings don't.

## ⚠️ Signature change

Per AGENTS.md. Parameters and arity unchanged.

```
MiscAPI.ListActions      ([]Action,  *http.Response, error)
                      -> ([]string,  *http.Response, error)

MiscAPI.ListPermissions  new
                      -> (map[string][]string, *http.Response, error)
```

The `Action` enum itself is untouched — it's still used by the role and binding models, where its staleness (no `READ`, `LOCK`, `PROMOTE`, `TEMPLATE`) is a separate question worth its own look.

## Verification

Live against **EE 1.3.35** and **EE 2.0.0-rc13**, both methods against both servers:

| | `ListActions` | `ListPermissions` |
|---|---|---|
| 1.3.35 | `[READ CREATE UPDATE DELETE]` ✅ | decode error — correct, wrong server line |
| 2.0.0-rc13 | 403 — correct, endpoint removed | 28 resources, no error ✅ |

`go build ./...` and `go vet` clean.

## Context

Found during a pre-GA sweep of the generated v1-compat surface against `kestra-ee.yml`, prompted by kestra-io/kestractl#117. Two other drifted endpoints turned up in the same audit — `FlowsAPI.UpdateTask` (`PATCH /flows/{ns}/{id}/{taskId}`) and `TriggersAPI.UpdateTrigger` (`PUT /triggers`) — both of which have **no** counterpart operation in the 2.0 spec at all and are called by nothing. Those are a separate cleanup PR; this one is the only drifted endpoint with a live consumer.

### DIFF
--- a/go-sdk/kestra_api_client/api_misc.go
+++ b/go-sdk/kestra_api_client/api_misc.go
@@ -350,12 +350,21 @@ func (r ApiListActionsRequest) GetTenant() string {
 	return r.tenant
 }
 
-func (r ApiListActionsRequest) Execute() ([]Action, *http.Response, error) {
+func (r ApiListActionsRequest) Execute() ([]string, *http.Response, error) {
 	return r.ApiService.ListActionsExecute(r)
 }
 
 /*
 ListActions Retrieve list of actions
+
+Kestra 1.x only: 2.0 removed /acls/actions when it replaced the flat action
+vocabulary with the per-resource model. Use ListPermissions against 2.x.
+
+The actions come back as plain strings rather than the Action enum, for the same
+reason as ListPermissions: this endpoint reports the server's own vocabulary, and
+a closed enum cannot represent a server that disagrees with it. Kestra 1.3
+returns READ, which the Action enum has never had, so every 1.3 call decoded
+into []Action failed even though the server answered 200.
 
 Actions are used to restrict possible operations for each permission. Each action must be one of the following: CREATE, READ, UPDATE, DELETE. Using permissions and actions together, you can control access to resources e.g. only allow a user to read a flow, but not update or delete it.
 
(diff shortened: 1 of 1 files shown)


AcevedoR:
Parking this as a draft — not a priority right now, so please don't spend review time on it.

Context is preserved above and the branch stays put. Short version of why it's worth picking up later: `kestractl server actions` is dead on Kestra 2.0 (403 on the removed `/acls/actions`), and `ListActions` has silently failed to decode on *every* 1.3 call for a while — the server returns `READ`, which the `Action` enum has never contained. kestractl has been masking that by recovering the list out of the error body.

Nothing depends on this, and no other work is blocked by it.

PR #413, opened 2026-09-04
feat: bind the Reusable Inputs endpoints across every SDK

brian-mulier-p:
Reusable Inputs (EE 2.0) was the last 2.0 feature with no SDK support — the reason kestra-io/kestra-ee#10757 had to add its Terraform support over raw HTTP. This binds all six endpoints in every SDK: paged list, get (with `revision`), revision history, `PUT` create-or-update from raw YAML (with `failIfExists`), delete, and the namespaces-with-blocks lookup.

- **Go** — `ReusableInputsAPI` plus the `ReusableInputs`/`ReusableInputsWithSource`/`PagedResultsReusableInputsWithSource` models and a `ReusableInputs()` accessor on `KestraClient`. Hand-written, like the rest of the Go API layer since #230.
- **Python** — `ReusableInputsApi`, the models and a `reusable_inputs` property on `KestraClient`, hand-written like the rest of the Python SDK since #237.
- **Java** — `ReusableInputsApi`, the models, and a `reusableInputs()` accessor, hand-written since #222.
- **JavaScript** — the only generated SDK; it picks the source split up from the spec sync below, so the code here is test coverage only.

Worth flagging for JS consumers: `list`, `get` and `delete` are only reachable from `@kestra-io/kestra-sdk/reusable-inputs-admin`, not the obvious `reusable-inputs` entry point. The Micronaut export numbers their operationIds (`list_4`, `get_6`, `delete_5`) because those method names collide across controllers, and the hey-api plugin routes suffixed ids into an `<Tag>Admin` module. So 3 of the 6 functions sit under a name implying instance-level admin scope. Tenants, Cases, AI, WorkerQueues and TenantAccess are affected the same way; the fix is unique method names in the controllers upstream.

The models follow the split kestra-io/kestra-ee#10774 introduces: `ReusableInputsWithSource` is what every method returns — the four authorable fields (`id`, `namespace`, `description`, `inputs`) plus the six the server owns (`revision`, `last`, `created`, `updated`, `deleted`, `source`) — and the paged wrapper follows its element type, the same shape as `Flow`/`FlowWithSource`. The authorable `ReusableInputs` is not a wire type and is gone: it only ever fed the block editor's schema, and the PUT body is raw YAML. TypeScript gets a flat object plus a `Writable` variant that strips the six read-only fields.

**Spec sync.** `kestra-ee.yml` is regenerated from `releases/v2.0.x` — EE and OSS `generateOpenapiSpec` plus the Enterprise-only route tagger, exactly as `update-openapi-spec.yml` does it — now that kestra-io/kestra-ee#10774 is merged and backported (kestra-io/kestra-ee#11001). That matters beyond this feature: the committed spec came from `develop`, which is now 2.1.0-SNAPSHOT, while this SDK releases on 2.0.x. Both lines expose the same 608 operations and differ in 13 hunks; only one has SDK reach, 2.0 declaring the export payloads `format: binary` rather than `byte`, so they type as `Blob | File` and the Apps/Dashboards export tests normalise the payload before asserting on it. `ReusableInputsWithSource` is a flat schema — the type does not extend the authorable one — and `ReusableInputs` is absent from the spec entirely: it was only ever the `allOf` parent, never a wire type, since the PUT body is raw YAML. The hand-written `ReusableInputs` models went with it.

**Enum drift fixed along the way.** The hand-written Go and Java `Type` enums stopped at `FORM`, while the spec has listed `ION` and `REUSABLE_INPUTS` for a while — including on `main`. That is reachable through these endpoints, since `ReusableInputsWithSource.inputs` is a list of `InputObject`: Go's `UnmarshalJSON` rejects an unlisted value, so a block with an `ION` input failed to decode and the whole response (or the whole page) was discarded; Java's `fromValue` fell through to `UNKNOWN_DEFAULT_OPEN_API` and wrote that back in place of the real type. Both enums now carry all 20 spec values, with a test per language. Python and JavaScript were already complete.

**Two knowingly-accepted consequences of the pin.** kestra-ee's `update-openapi-spec.yml` runs on every push to `develop` and copies its spec wholesale over `kestra-ee.yml`, so the next bot run will revert this pin — that is accepted rather than guarded: the pin only has to hold until the 2.0.1 release is cut, and the test layer absorbs the difference (`asText` normalises the `binary`/`byte` export flip). Please don't re-pin it reflexively when the bot PR lands. Relatedly, `compute-version` resolves `develop` for a `main`-based PR, so CI generates the JS client from the 2.0 spec and e2e-tests it against a 2.1 server; the 2.0 client was instead verified by hand against an EE 2.0 instance, as described below.

### ⚠️ Signature change

No method signature changed — all six methods per language are new.

Two **Python exports are removed**: `kestrapy.ReusableInputs` and `kestrapy.PagedResultsReusableInputs`, along with `kestrapy/models/reusable_inputs.py` and `paged_results_reusable_inputs.py`. Both ship in every published 2.0.0 python artifact (`v2.0.0-rc1-python` through `v2.0.0-python`), so `from kestrapy import ReusableInputs` starts raising `ImportError`. Nothing could have used them meaningfully — no endpoint returns the authorable type — but downstream consumers should know before they upgrade. Go and Java never shipped an equivalent (no `go-sdk/v2.0.0*` or `*-java` tag contains one).

**Verification:** built an EE instance from kestra-io/kestra-ee#10774 (Postgres, tenant `demo`) and round-tripped a block through each hand-written layer against it — paged list, get by revision, revision history, create-or-update from YAML with `failIfExists`, delete and the namespaces lookup — asserting `revision` bumps 1 → 2, `last` flips to false on a pinned revision and `source` comes back verbatim: Go 12/12, Java 12/12, Python green. On the 2.0 spec: `./generate-sdks.sh javascript` regenerates clean and leaves no tracked diff, `npm run check:types` passes, `go build ./...` and `go vet ./...` are clean, Java `compileJava` and `:java-sdk-test:compileTestJava` compile against a freshly published `local-ci-version`, the Python SDK imports with 542 tests collected and `validate_doc_examples.py --check` green, and all four README snippet checks pass.


jymaire:
## Code review

The plumbing checks out — paths, tenant prefixing, query-param builders, YAML content-type handling, `invoke`/`doJSON`/`_json_request` arities, the JS generated function names, and the test helpers all match `kestra-ee.yml` and the existing conventions in each SDK.

Four findings. Three share one root cause: `revision`, `deleted` and `last` are modelled as if kestra-ee#10774 had already shipped, but none of them are in the checked-in `ReusableInputs` schema (required is only `namespace`/`id`/`inputs`).

### 1. `go-sdk/kestra_api_client/model_reusable_inputs.go:32` — Go hard-fails if `revision` is absent

`Revision int32` is a non-pointer field *and* is listed in `UnmarshalJSON`'s `requiredProperties` (line 423). If any EE response omits `revision`, **every** Go method returns `no value given for required property revision` and discards the whole response — including the entire page in `ListReusableInputs`. Java (Jackson, no validation) and JS (field absent from the generated types) degrade gracefully; Go is the only SDK that breaks. Unverified against a live server, per the PR description.

Suggested: `Revision *int32`, and drop it from `requiredProperties` until the spec declares it required.

### 2. `python/python-sdk/kestrapy/models/reusable_inputs.py:36` — unflagged breaking change

`revision: StrictInt` with no default means `ReusableInputs(namespace=..., id=..., inputs=[...])` — valid today, and still the shape the published spec describes — now raises `ValidationError`. `from_dict()` on a payload without `revision` also raises; `BaseApi._deserialize` then silently falls back to `model_construct`, after which `obj.revision` raises `AttributeError` instead of returning `None`.

The PR description says "Signature change: None"; per `AGENTS.md` this one needs the `### ⚠️ Signature change` note. Suggested: `Optional[StrictInt] = None`, matching the treatment of `source`/`last`/`deleted`.

### 3. `javascript/test_javascript_sdk/ReusableInputsApi.spec.ts:15` — self-swallowing throw

The `throw new Error(...)` sits inside the `try`, so its own `catch` eats it: `status` resolves to `undefined` and the test fails with the misleading `expected undefined to be 409` instead of the intended message. `CasesApi.spec.ts:72` has the pattern that avoids this (`statusOf` returning `0` on success) — either move the throw after the `try`/`catch`, or return a sentinel.

### 4. `go-sdk/test/api_reusable_inputs_test.go:43` and `:148` — vacuous assertions

`require.False(t, created.GetDeleted())` and `require.False(t, first.GetLast(), ...)` pass whether the server sends `false` or omits the field entirely — which is exactly the uncertain case here, since `deleted`/`last` aren't in the checked-in spec. Use `GetDeletedOk()` / `HasDeleted()` (resp. `GetLastOk()`) to assert presence *and* value. The Python equivalents (`assert created.deleted is False`) are already strict; only Go is lenient.

---

Making `revision` optional in Go and Python, plus tightening the Go presence assertions, resolves 1, 2 and 4 together.

jymaire (approved):

PR #414, opened 2026-09-07
fix(java)!: align the Java SDK with the Kestra 2.0 API surface

jymaire:
Targets `releases/v2.0.x` only — the branch cut at 90467c89, before #389 brings the 2.1 spec onto `main`.

First of four language-by-language PRs (Java, then Python, Go, JS).

## Why

`main`'s CI validates against `develop` (now `2.1.0-SNAPSHOT`), so the 2.0 GA delta was never exercised. Diffing the vendored spec against the `kestra-ee@v2.0.0` spec turned up three Java-visible items.

## What

**1. `BulkErrorResponse` removed**

The schema is gone from kestra-ee at the `v2.0.0` tag — `ProblemDetail` (RFC 9457) is now the only error model. The 2.0 release notes call this out for libraries: a body field like `message` is now `title`/`detail`. The model could only ever fail to deserialize, so it and its doc are deleted and de-registered from `.openapi-generator/FILES` and `README.md`.

**2. `FlowsApi.getConcurrencyLimit` bound**

`ConcurrencyLimitController#getConcurrencyLimit` (`ConcurrencyLimitController.java:39` at `v2.0.0`) had no Java binding.

```java
public ConcurrencyLimit getConcurrencyLimit(String namespace, String flowId, String tenant)
```

**3. `UsersApi.deleteUsersByIds` bound**

`IAMUserController` `@Delete(uri = "by-ids")` (`IAMUserController.java:239`) had no Java binding. Adds the request model `IAMUserControllerApiDeleteUsersRequest` (`List<String> ids`).

```java
public BulkResponse deleteUsersByIds(IAMUserControllerApiDeleteUsersRequest request)
```

The whole request is rejected server-side if any id does not resolve to a user, or if it contains the caller's own id.

### ⚠️ Signature change

None. Both API changes are new methods; no existing public method's parameters, types or arity changed.

The one removal is a **model class**, not a method: `io.kestra.sdk.model.BulkErrorResponse` is deleted. Nothing in the SDK referenced it, but it was public API, hence the `!` in the commit subject.

## Deliberately not in this PR

`java/java-sdk/api/openapi.yaml` still carries the `2.0.0-SNAPSHOT` snapshot and its `BulkErrorResponse` `$ref`s. It is a vendored spec artifact — re-synced from a v2.0.0 spec export, not edited by hand. That re-sync also covers the JS `exportAuditLogs` filters migration, which is the sharper remaining risk (the removed query params are silently ignored by Micronaut, so 2.0 answers 200 with an unfiltered CSV).

## Verification

- `./gradlew clean compileJava` — green
- no `BulkErrorResponse` reference left in Java sources, docs, FILES or README
- every upstream reference checked at the `v2.0.0` tag, not `develop`

Test suites are run separately against a pinned v2.0.0 image; the release workflows do not re-run tests at tag time, so that green run is the gate.


jymaire:
Closing — the 2.0 release-branch approach is being reverted. The work is preserved at 156cc32d (Java-only) and 970c3ea0 (all four SDKs) if we pick it back up.

Issue #331, opened 2026-07-08
[EPIC] Increase coverage across all client SDKs


jymaire:
## Summary

Coverage of the Kestra HTTP API is uneven across our four client SDKs, and only one of them measures it in CI. This EPIC groups the work to raise coverage for **all four** SDKs and to make every SDK's coverage **continuously measured** on every PR.

Two distinct axes are in play, and which one matters depends on how the SDK is built:

- **API-surface coverage** — does the SDK wrap the endpoint at all? This is the gap for the **hand-written** Go/Python/Java SDKs: last reliable measurement (2026-05-27) put **Python ≈ 59%**, **Go ≈ 57%**, **Java ≈ 54%** of the **~464 `/api/v1` endpoints**.
- **Function test coverage** — is each wrapped function exercised by a test? This is the gap for the **generated** JavaScript SDK, which already wraps the full surface: **71.1% (308 / 433 functions)**, with 125 functions untested.

## Motivation

- A user who picks a hand-written SDK hits a wall the moment they need an endpoint that isn't wrapped, and drops to raw HTTP — defeating the purpose of the SDK.
- Coverage is uneven across languages: the same automation is possible in one SDK and impossible in another.
- The surface keeps growing (~464 endpoints now vs ~441 in May). Only JS has a coverage gate today, so the other three silently regress whenever the backend adds an endpoint.

## Prerequisite — measure coverage in CI for every SDK

**JavaScript already does this well:** `.github/scripts/coverage-comment.mjs` (wired into `.github/workflows/javascript-sdk.yml`) reads the istanbul `coverage-final.json` and posts a sticky PR comment listing every untested function in `openapi/sdk/`. This is the model to generalize.

- [ ] Extend JS-style coverage reporting to **Go, Python, and Java** so each PR shows the current number and blocks regressions.
- [ ] For the hand-written SDKs, also revive the **API-surface** measurement. A script already exists on the unmerged `tech/add-coverage` branch (`scripts/api_coverage.py`, commit `519a2c1a`) — it derives the surface from backend controllers (`kestra` + `kestra-ee`) rather than the OpenAPI spec, which is being removed. It has drifted from the current SDK layout (currently reports 0% for Python/Go, "api dir not found" for Java) and must be fixed and merged before it is trustworthy.

## API surface at a glance (`develop`, ~464 `/api/v1` endpoints)

Largest resource groups by endpoint count — a map of where the surface work concentrates for the hand-written SDKs:

[...]

PR #429, opened 2026-09-15
feat(java): reach ≥90% SDK API coverage — policies, mcp-servers, metrics, notifications, case-templates, banners, kill-switches (#422)

### LINKED ISSUE #422: Increase Java SDK API coverage (surface ~54% → ≥90%) + fix /actions/ URL bug

jymaire:
Closes #422 (parent EPIC #331).

Raises Java SDK **API-surface coverage to ~92.9%** of the **607** `/api/v1` operations in `kestra-ee.yml`, clearing the ≥90% target.

## What changed

New/extended hand-written API classes, all wired into `KestraClient`:

| API | Endpoints | Scope |
|---|---|---|
| `ScimApi` | 19 | `/{tenant}/integrations/{integration}/scim/v2/**` — Users, Groups, ResourceTypes, Schemas, ServiceProviderConfig (SCIM JSON) |
| `PoliciesApi` | 21 | tenant `/{tenant}/policies/**` + instance `/instance/policies/**` — CRUD, evaluate, export, export-by-ids, import, validate |
| `McpServersApi` | 7 | `/{tenant}/mcp-servers/**` — CRUD, tools, toggle |
| `MetricsApi` | 7 | `/{tenant}/metrics/**` — names, tasks, aggregates, by-execution + `/instance/metrics/{serviceType}` |
| `NotificationsApi` | 6 | `/notifications/**` — since, history, unread-count, read/unread, read-all |
| `CaseTemplatesApi` | 5 | `/{tenant}/case-templates/**` — CRUD + search |
| `BannersApi` | 4 | `/banners/**` — CRUD + search (instance-owner) |
| `KillSwitchesApi` | 4 | `/kill-switches/**` — CRUD + search (instance-owner) |
| `AppsApi` (extended) | 4 | `states`, open-by-uid, `preview`, `view/{id}/file/download` |

## Review fixes (from the #429 review)

1. **`downloadFileFromAppExecution` produces mismatch** — it sent `Accept: application/octet-stream`, but the route declares `application/json`, so the request was rejected at routing. Switched to `JSON`; `ApiClient.deserialize` short-circuits on `byte[]` before the mime check, so the bytes still come back. The missing-app test now asserts **404** (the controller's `notFound` branch — there is no 403 branch) and a positive path against a deployed app was added.
2. **Typed models instead of `Map<String,Object>`** where the class already ships in `io.kestra.sdk.model`: `PoliciesApi` validate → `ValidateConstraintViolation`, delete-by-ids → `BulkResponse`; `BannersApi` → `Banner`; `KillSwitchesApi` → `KillSwitch`; `MetricsApi` search → `PagedResultsMetricEntry`, aggregates → `MetricAggregations`. `date-time` params are now `OffsetDateTime` (metrics aggregate start/end, notifications `since`), matching `ExecutionsApi.createExecution`.
3. **Coverage denominator corrected** — the spec has **607** `/api/v1` operations, not 587. The two remaining instance/config gaps the review flagged are now closed: `GET /instance/metrics/{serviceType}` (`MetricsApi.instanceServiceMetrics`) and the full **SCIM v2** family (`ScimApi`), which together take the branch from ~89.8% to ~92.9%.
4. **Test hygiene** — every `create()` now sits inside the `try`, so a failed post-create assertion can no longer leave an instance policy / banner / kill switch / MCP server live and cascade into later tests; cleanup goes through a `deleteQuietly` helper.

## Out of scope (the remaining ~7%)
Auth/session routes (`login`, `logout`, `setup`, password reset, `configs`, `request-code`, `basicAuth`), SSE `follow`/`streams` endpoints (some already wrapped via `sseFlux`, invisible to a verb-regex), and multipart app `dispatch`.

## Signature changes
All within methods **introduced in this PR** (no released signature dropped): the return-type and `date-time`→`OffsetDateTime` refinements listed under review-fix #2. Every change is additive at the class/method level (new classes, new methods, new `KestraClient` accessors).

## Tests
New live test classes assert real response values (round-trip create/search/get/update/delete where the body schema is known; paged-envelope / discovery-document shape and 404 binding-guards elsewhere). `ScimApiTest` provisions a real (auto-enabled) SCIM security integration — the CI image licenses `FEATURE_SCIM` — to exercise the read-only discovery/query endpoints. Main + test sources compile (`compileJava`, `compileTestJava`). The live suite needs a running Kestra CI image; it has **not** been run here — please confirm green in CI.


Malaydewangan09:
Reviewed against `kestra-ee.yml`, the EE controllers and the CI run. 43 of the 44 new endpoints map correctly, and the round-trip tests assert real values rather than just shapes. Three things to fix.

### 1. `downloadFileFromAppExecution` never reaches the controller

`AppsApi.java:310` sends `Accept: application/octet-stream`, but the route declares `application/json` (`@Get(uri = "/view/{id}/file/download")`, no `produces`). The request is rejected at routing.

The test is green for the wrong reason. `checkAccessLevelAndGet` returns empty for a missing app and the method then returns `notFound()`, there is no 403 branch. `openApp` goes through the same helper and your own test observes 404 there. And `PluginsApiTest.java:218-223` on main already records this exact 403 for the only other endpoint with the same produces mismatch.

Fix is `JSON` instead of `OCTET_STREAM`. `ApiClient.deserialize` short-circuits on `byte[].class` before it looks at the mime type, so the bytes still come back. Then flip the test to 404 and add a happy path, `AppsApiTest` already has the fixtures.

### 2. Typed models are being bypassed

`NamespacesApi` on main returns `BulkResponse` and `ValidateConstraintViolation` for the namespace twins of these same controller methods. `PoliciesApi.java:94,114,215,233` returns `Map<String, Object>`. `Banner`, `KillSwitch`, `MetricAggregations` and `PagedResultsMetricEntry` also ship in `io.kestra.sdk.model` with zero references anywhere, so `BannersApi`, `KillSwitchesApi` and `MetricsApi` could use them directly. `Map<String, Object>` is fine where no model exists, less so when the class is sitting right there.

### 3. The coverage number

The spec has 607 `/api/v1` operations, not 587. Parsing every `invoke`/`sseFlux` call site plus the per-file path and verb helpers gives main at 80.2% and this branch at 89.6%. Your +9.7pp delta is basically right (+9.4pp by my count), so the work holds, but the absolute lands just under the 90% gate.

Two gaps missing from the out-of-scope list:

- The whole SCIM v2 family, 16 ops under `/{tenant}/integrations/{integration}/scim/v2/**`. Nothing in the SDK references `scim`. Those alone would take it to 92.3% and genuinely clear the target.
- `GET /api/v1/instance/metrics/{serviceType}` is still unwrapped, so `instance/policies` was not the only instance gap. Natural fit for the new `MetricsApi`.

### Test hygiene

`PoliciesApiTest.java:179` asserts before the `try`/`finally`. If it fails, an instance policy that BLOCKs `io.kestra.plugin.core.log.Log` stays live for the rest of the suite and cascades into every test that runs a Log flow. Same shape at 119, 150, 204 and in `McpServersApiTest`, `BannersApiTest`, `KillSwitchesApiTest`. Move the create inside the `try`.

Minor: `MetricsApi.java:84,100` and `NotificationsApi.java:31` take `date-time` params as `String`, where `ExecutionsApi.createExecution` uses `OffsetDateTime`.

Security is clean. Every path segment goes through `esc()` and is URL-encoded, so no traversal. Additive-only and the no-signature-change claim both check out.

jymaire:
Thanks — all confirmed against the EE controllers and fixed in 61ad100f.

**1. `downloadFileFromAppExecution` mismatch.** Right on both counts. `AppsController.downloadFileFromAppExecution` has no `produces` (defaults to JSON) and `checkAccessLevelAndGet(id, ACCESS_FILES)` → empty → `notFound()`, with no 403 branch. Switched Accept to `JSON` (bytes still return via the `byte[]` short-circuit in `deserialize`), flipped the test to **404**, and added a positive path: a deployed app with no Outputs/TaskOutputs block hits `hasAccessToExecutionFile(...) == false` → controller `badRequest` (400), proving the route is reached.

**2. Typed models.** Done — `PoliciesApi` validate → `ValidateConstraintViolation`, delete-by-ids → `BulkResponse` (matches `PolicyController`); `Banner`, `KillSwitch`, `PagedResultsMetricEntry`, `MetricAggregations` now used directly. `date-time` params → `OffsetDateTime` (metrics aggregates, `notifications/since`).

**3. Coverage.** Confirmed 607 `/api/v1` ops (not 587). Wrapped the two gaps you named — `GET /instance/metrics/{serviceType}` (`MetricsApi.instanceServiceMetrics`) and the full **SCIM v2** family (new `ScimApi`, 19 ops). That takes the branch from ~89.8% to ~92.9%, genuinely clearing the gate. Description updated with the corrected denominator.

**Test hygiene.** Every `create()` moved inside the `try`; cleanup via a `deleteQuietly` helper (following `TenantsApiTest`), so a failed post-create assertion no longer leaves a BLOCKing policy / banner / kill switch / MCP server live.

Main + test sources compile; live suite still needs CI to confirm green.

Malaydewangan09:
Re-reviewed `61ad100f` end to end against `kestra-ee.yml`, the EE controllers and the CI run.

The download fix is correct and proven: both `downloadFileFromAppExecution` tests passed live, 404 for a missing app and 400 for an app with no Outputs block. `ScimApi` matches the spec exactly on all 19 ops, including the odd `ScimUser` body on `POST .../Groups` (that is the server's own quirk, worth a one-line comment so it does not read as a copy/paste slip). `application/scim+json` is covered by `ApiClient.isJsonMime`, and `parameterToString` formats the new `OffsetDateTime` params properly. Coverage recount confirms your number: **564/607 = 92.9%**.

CI is red though, 86 failures, and one of them is mine to own.

### `KillSwitch` cannot round-trip, and it takes the suite down with it

`io.kestra.ee.killswitch.KillSwitch` declares `LocalDateTime startDate/endDate`. The generated `io.kestra.sdk.model.KillSwitch` declares `OffsetDateTime`, because the spec flattens `LocalDateTime` to `format: date-time` and loses the no-offset part. So the server echoes `"2999-01-01T00:00:00"`, `RFC3339InstantDeserializer` rejects it, and `createKillSwitch` throws `ApiException{code=0}` on a **200 response**:

```
InvalidFormatException: Cannot deserialize value of type `java.time.OffsetDateTime`
  from String "2999-01-01T00:00:00"   (through reference chain: KillSwitch["startDate"])
  at KillSwitchesApi.createKillSwitch(KillSwitchesApi.java:30)
```

My earlier "use the typed model" note was wrong for this one class. `Banner`, `MetricEntry` and `MetricAggregation` are all `Instant` server-side so `OffsetDateTime` is fine there. `KillSwitch` is the single outlier.

Then the cascade. The switch *was* created server-side, but `id` stays null, so `if (id != null) deleteQuietly(id)` is a no-op and a live `KILL` switch survives. Everything after it breaks: 81 more failures across `KvApiTest`, `LogsApiTest`, `MetricsApiTest`, `NamespacesApiTest`, `PluginsApiTest`, `TestSuitesApiTest`, `TriggersApiTest` and `V2FeaturesApiTest`, all 500s on `POST /flows`, `/plugins/*`, `/logs/search` and `/tests/search`.

Fix the model rather than reverting to `Map`. `LocalDateTime` matches the EE record, `RFC3339JavaTimeModule` only overrides `Instant`, `OffsetDateTime` and `ZonedDateTime` so `LocalDateTime` falls through to the stock `JavaTimeModule` deserializer, and `WRITE_DATES_AS_TIMESTAMPS` is already disabled so it serializes back as `"2999-01-01T00:00:00"`. Works both directions. The header says "do not edit", but `generate-sdks.sh` refuses to run for Java so it will not be clobbered. The spec defect is worth its own upstream issue.

Note that `deleteQuietly` alone cannot save this class of failure, and a name-prefix sweep would not either, since `searchKillSwitches()` returns `List<KillSwitch>` and fails the same way. The model fix is load-bearing. Given the blast radius, consider `PASS` instead of `KILL` for the fixture too.

### `ScimApiTest` never gets past setup

All 4 tests die at `createScimIntegration` (line 37) with 422 `createSecurityIntegration.body.description: must not be null`. `CreateSecurityIntegrationRequest` requires `description`, `name` and `type`. The setup call also sits outside the `try`, so `deleteQuietly` cannot run, which is the same shape you just fixed in the other classes.

Two more things while you are in there. Each test creates its own integration, and a SCIM create also provisions a service account plus API token that `deleteSecurityIntegration` does not cascade, so that is 4 orphan service accounts per run. `@BeforeAll` / `@AfterAll` with one integration would cut it to one. And `getUser_unknownId_reachesController` only asserts `code != 403`, which a 500 satisfies. The spec documents 404, so assert that.

Bigger picture: 15 of the 19 SCIM ops have no test at all. `getResourceType`, `getSchema` (where a `urn:...` path segment gets percent-encoded, untested), `findUsers`, `findGroups`, and every create/update/patch/delete are untouched. The class javadoc says the writes "are exercised for binding/coverage only", but they are not exercised at all.

### Smaller

`instanceServiceMetrics` is right even though it looks wrong. `ClusterController` declares `@Get("/metrics/{serviceType}")` with `@QueryValue ServiceType serviceType`, so the server genuinely needs both the path segment and the query param. Worth a comment saying so, otherwise someone will "clean it up" and break it.

`instanceServiceMetrics_forWebserver_returnsMetrics` uses `allSatisfy` on a list that could be empty, which passes vacuously and does not match the comment's claim. Add `isNotEmpty()`. That test never actually ran, it was in the cascade.

`BannersApiTest` dropped the post-delete `noneSatisfy` check, so delete is no longer verified.

Malaydewangan09 (approved):

PR #442, opened 2026-09-18
fix(go): tenant-scope AiAPI.GenerateFlow / ListAiProviders (#436)

### LINKED ISSUE #436: Go SDK: AiAPI.GenerateFlow / ListAiProviders hardcode the 'main' tenant
### Summary

`AiAPI.GenerateFlow` and `AiAPI.ListAiProviders` in the Go SDK build their path with `superadminPath("main", "ai", ...)`, hardcoding the tenant as `main`:

```go
// go-sdk/kestra_api_client/ai_api.go
func (a *AiAPI) GenerateFlow(ctx context.Context, body interface{}) (string, error) {
    return a.postForYAML(ctx, "POST", superadminPath("main", "ai", "generate", "flow"), body)
}
func (a *AiAPI) ListAiProviders(ctx context.Context) ([]map[string]interface{}, error) {
    return doJSON[[]map[string]interface{}](&a.baseAPI, ctx, "GET", superadminPath("main", "ai", "providers"), nil, nil)
}
```

But the EE AI controller is mounted at `@Controller("/api/v1/{tenant}/ai/")`, and the spec has `/api/v1/{tenant}/ai/generate/flow` and `/api/v1/{tenant}/ai/providers`. Both routes are tenant-scoped, and the sibling generators (`GenerateApp`/`GenerateDashboard`/`GenerateTest`) already take a `tenant` correctly.

### Impact

On a single-tenant `main` install it works; on any other tenant it silently targets the wrong tenant, and the caller has no way to override it (neither method accepts a `tenant`).

### Fix

Add a `tenant string` parameter to both methods and use `tenantPath(tenant, "ai", ...)`, matching the other four generators.

### Context

Found during the Python SDK #421 coverage review (#433) — the Python wrappers were ported 1:1 from this Go reference and inherited the same bug. Fixed on the Python side in that PR; filing here so the two SDKs reconverge.

jymaire:
Fixes #436.

`AiAPI.GenerateFlow` and `AiAPI.ListAiProviders` built their path with `superadminPath("main", "ai", ...)`, hardcoding the tenant as `main`. But the EE AI controller is mounted at `@Controller("/api/v1/{tenant}/ai/")` and the spec exposes both `/api/v1/{tenant}/ai/generate/flow` and `/api/v1/{tenant}/ai/providers` as tenant-scoped. On any non-`main` tenant these silently hit the wrong tenant, and the caller had no way to override it.

## Changes

- `GenerateFlow` and `ListAiProviders` now take a `tenant` parameter and use `tenantPath(...)`, matching the four sibling generators (`GenerateApp`/`GenerateDashboard`/`GenerateTest`) and the already-fixed Python wrappers (#433 — this reconverges the two SDKs).
- Corrected the `AiAPI` type doc comment and both method doc comments, which claimed some AI routes lived under `/api/v1/main/ai/*`; every AI route is tenant-scoped.
- Updated the one in-repo caller (`go-sdk/test/api_ai_test.go`).

`superadminPath` remains — it's still used by many other genuinely instance-scoped APIs.

## ⚠️ Signature change

```
GenerateFlow(ctx, body)   → GenerateFlow(ctx, tenant, body)
ListAiProviders(ctx)      → ListAiProviders(ctx, tenant)
```

This is a deliberate, correctness-driven change (per `AGENTS.md` → *Signature Changes — Notify, Don't Block*). Blast radius is minimal: the only in-repo caller is the AI test (updated here), and the primary downstream consumer `kestractl` uses neither method and is pinned to `go-sdk/v2.0.1`.

## Testing

`go build ./...` and `go vet ./...` pass. The AI test self-skips without a configured provider (`skipIfGated`), so no live coverage change.


fdelbrayelle:
Reviewed. Diff is clean, no bugs found.

**On "shouldn't main be the default tenant if absent":** no — checked `tenantPath` and every sibling generator (`GenerateApp`/`GenerateDashboard`/`GenerateTest`, `ListThreads`, etc.): none default an empty tenant to `"main"`, all require it explicit, `tenantPath` itself has no fallback. Adding a default here would be inconsistent with the rest of the SDK, not a fix. This PR correctly follows that existing convention.

Other points checked:
- Doc comments accurately updated (no more `/api/v1/main/ai/*` claims).
- In-repo caller (`api_ai_test.go`) updated to pass `MAIN_TENANT`.
- New `api_ai_tenant_path_test.go` is a good addition — hermetic mock-server test that actually exercises the fixed byte, since the real `TestAiAPI` self-skips without a configured AI provider.
- Signature-change note present per AGENTS.md, blast radius correctly scoped (kestractl pinned to v2.0.1, doesn't call these two methods).

LGTM.

fdelbrayelle (approved):

PR #443, opened 2026-09-18
fix(java): restore flow inputs on createExecution (#360)

### LINKED ISSUE #360: `createExecution` no longer accepts flow inputs (regression since 1.0.11)
Since 1.3.x (checked on 1.3.1 and 1.3.2), `ExecutionsApi.createExecution(...)` has a single signature with no inputs parameter, and sends a `null` body despite the endpoint being `multipart/form-data`. In 1.0.11 there was an overload taking `HashMap<String, Object> inputs`.

No other method on `ExecutionsApi` lets you pass inputs, so you can't create a parameterized execution via the SDK anymore — even though `POST /executions/{namespace}/{flowId}` accepts inputs as form-data. This blocks upgrading past 1.0.11.

Repro: compare `javap` of `ExecutionsApi.createExecution` between 1.0.11 and 1.3.2.

jymaire:
## What

Restores the ability to create parameterized executions from the Java SDK. Since 1.0.11, `ExecutionsApi.createExecution(...)` declared `multipart/form-data` but had **no inputs parameter** and sent an empty body, so there was no way to pass flow inputs when creating an execution — blocking upgrades past 1.0.11.

## Changes

- **`createExecution`** — new additive overload `createExecution(..., Map<String, Object> inputs)` (last param). The existing 9-arg method delegates to it with `null`, so existing callers are unaffected.
- **`replayExecutionWithInputs`** — had the identical defect (declared multipart, named "with inputs", took none). Same additive overload added.
- **`toFormParams` helper** — inputs are encoded per flow-input type, mirroring the already-working Python SDK:
  - `File` / `byte[]` → binary part (for `FILE` inputs)
  - `String` → verbatim
  - anything else (`Boolean`, `Integer`, `Map`, `List`, …) → JSON-encoded
  - `null` values skipped (so the flow's declared `defaults` apply)
- **`resumeExecution`, `validateNewExecutionInputs`, `validateResumeExecutionInputs`** — routed through the same helper. They already accepted `inputs`, but serialized non-string values via `Object#toString` (`ApiClient.parameterToString` → `String.valueOf`), which sends invalid JSON for `JSON`/`ARRAY` inputs (`{k=v}` instead of `{"k":"v"}`). Now fixed for those endpoints too.

## Tests

Added to `ExecutionsApiTest`:
- `createExecution_withStringInput` — reads the value back from the execution (`required: true` input, so it only passes if inputs are forwarded).
- `createExecution_withTypedInputs` — `BOOL` + structured `JSON` round-trip; guards against the `String.valueOf` mis-serialization.
- `createExecution_withFileInput` — `byte[]` FILE input resolves to a storage URI.
- `createExecution_withNullInputs_stillWorks` — the no-inputs path still sends a valid empty multipart body.
- `replayExecutionWithInputs_overridesInput` — replay carries the overridden input value.

> Note: these are live integration tests (require a running Kestra); they were compiled here but the suite is run by the maintainer.

### ⚠️ Signature change

No existing signature changed, so nothing downstream recompiles. Two kinds of change:

**Additive overloads:**
- `createExecution(String, String, String, List<String>, Boolean, Integer, OffsetDateTime, String, ExecutionKind)` → **plus** `createExecution(…, Map<String, Object> inputs)`
- `replayExecutionWithInputs(String, String, String, Integer, String)` → **plus** `replayExecutionWithInputs(…, Map<String, Object> inputs)`

**Wire-format change (no signature change) for `resumeExecution`, `validateNewExecutionInputs`, `validateResumeExecutionInputs`** — now routed through `toFormParams`, which changes what existing callers put on the wire:
- a `List` value was comma-joined by `parameterToString` (`"a,b"`) and is now JSON (`["a","b"]`)
- a `null` value was sent as an empty-string part and is now omitted, so the flow's declared `default` applies

Both are fixes (they make `JSON`/`ARRAY` inputs parseable server-side), but they are behavior changes for existing callers, noted here so downstream consumers can catch anything unexpected.

## Scope notes

- **Java-only**, matching the issue (the `javap` / 1.0.11 evidence is all Java). Python already supports `inputs` end to end with passing tests.
- **Go is impacted in serialization, not in surface.** The Go SDK's fluent client (used by `kestractl`) already exposes `.FormData(inputs)`, so the surface is fine — but `api_executions.go` ends its form-encoding type switch with `default: localVarFormParams.Add(key, fmt.Sprint(v))`, so a `map[string]any` JSON input goes out as `map[count:2]` and dies on the same `JacksonMapper.toObject` this PR fixes for Java, and `[]interface{}` is added as repeated fields rather than JSON. Anyone passing a structured `JSON`/`ARRAY` input through kestractl today hits this exact bug. It is the **same defect**, not a tidy-up, and is left as a follow-up since this PR is scoped Java-only per the issue.

closes: https://github.com/kestra-io/client-sdk/issues/360


Malaydewangan09:
Reviewed `739bdfe3` end to end against the spec, `ApiClient`'s multipart builder, Kestra's `FlowInputOutput`, and the CI run.

The diagnosis and the fix are right, and the tests are the good kind: real value round-trips, not shape checks. But CI is red, and the one failure is real.

### `createExecution_withFileInput` fails: `byte[]` parts are sent without a filename

One failure out of 643 tests, and it is the FILE input path:

```
422 Invalid value for input `attachment`.
    Cause: The file 'file-input-content' does not exist.
```

The filename in that message is the *content* of the byte array. So the server bound `attachment` as a plain text attribute whose value happened to be the bytes, then tried to resolve that string as a file path.

Root cause is in `ApiClient.java:858`:

```java
} else if (value instanceof byte[]) {
    multiPartBuilder.addBinaryBody(paramEntry.getKey(), (byte[]) value);
}
```

That overload emits a part with no `filename` in its Content-Disposition. Micronaut binds a filename-less part as a `String` attribute rather than a `CompletedFileUpload`, and Kestra's FILE input resolution then treats the string as a path. The `File` branch one line above does not have this problem, because `addBinaryBody(name, File)` derives the filename from the file.

Fix is the four-arg overload:

```java
multiPartBuilder.addBinaryBody(paramEntry.getKey(), (byte[]) value,
        ContentType.DEFAULT_BINARY, paramEntry.getKey());
```

Blast radius is zero. Every existing form param in the Java SDK is a `File` or a `String` (`AppsApi`, `CasesApi`, `FilesApi`, `FlowsApi`, `InstanceApi`, `TenantsApi`), so that `byte[]` branch is dead code today and this PR is its first caller. It also brings Java in line with Python, whose `_multipart_upload` always supplies `(file_name, content)`.

Worth adding a `File`-valued FILE-input test too. That path should already work, but nothing covers it, and it is the other half of what `toFormParams` promises.

### The rest of it checks out

I verified rather than assumed:

- The regression is real as described. On `main`, `createExecution` calls the 8-arg `invoke(...)` with no form params, so `BaseApi` passes `new HashMap<>()` and the multipart body was always empty.
- The encoding matches what the server actually wants. `FlowInputOutput` resolves `case JSON -> ... JacksonMapper.toObject(current.toString())`, so `String.valueOf(Map)` producing `{count=2}` could never parse and `writeValueAsString` producing `{"count":2}` does. `createExecution_withTypedInputs` is precisely the right regression guard and it passed.
- `new ApiException(msg, e, 0, null)` resolves unambiguously: the other four-arg overload is `(String, int, Map, String)` and a `JsonProcessingException` cannot bind to `int`.
- Backward compatibility holds. The nine-arg delegates with `null`, `toFormParams(null)` returns an empty map, and the body is the same empty multipart as before. `createExecution_withNullInputs_stillWorks` passed.
- Compiles clean locally.

### A wire-format change worth a line in the description

Routing `resumeExecution`, `validateNewExecutionInputs` and `validateResumeExecutionInputs` through `toFormParams` changes what existing callers put on the wire, without changing any signature:

- a `List` value was comma-joined by `parameterToString` ("a,b") and is now JSON (`["a","b"]`)
- a `null` value was sent as an empty-string part and is now omitted, so the flow's declared default applies

Both are fixes, and I would not gate on them. But they are exactly the kind of change the `AGENTS.md` signature-change note exists to surface for downstream consumers, so they belong under that heading alongside the two additive overloads.

### Go is impacted, just not in the signature

`api_executions.go:216` ends its type switch with:

```go
default:
    localVarFormParams.Add(key, fmt.Sprint(v))
```

So a `map[string]any` JSON input goes out as `map[count:2]` and dies on the same `JacksonMapper.toObject` this PR is fixing in Java, and `[]interface{}` is added as repeated fields rather than JSON. "kestractl is not impacted" is true of the surface (`.FormData` exists) but not of the serialization: anyone passing a structured JSON or ARRAY input through kestractl today hits this bug. Worth saying so plainly in the description so the follow-up gets sized as the same defect rather than a tidy-up.

Malaydewangan09 (approved):

jymaire:
Thanks — spot on, all four addressed in `283c0f36`.

**The failure (root cause confirmed).** Reproduced the exact 422 on CI (`The file 'file-input-content' does not exist` — the filename was the byte content), and it's the filename-less `addBinaryBody` you pointed at. Applied your four-arg fix:

```java
multiPartBuilder.addBinaryBody(paramEntry.getKey(), (byte[]) value,
        ContentType.DEFAULT_BINARY, paramEntry.getKey());
```

Agreed on blast radius: every other Java form param is a `File` or `String`, so that `byte[]` branch was dead until this PR — and this brings it in line with Python's `(file_name, content)`.

**`File`-valued FILE-input test.** Added `createExecution_withFileInput_fromFile` next to the `byte[]` one, covering the other half of `toFormParams` (the `File` branch, one line above the bug).

**Wire-format note.** Added to the description under `⚠️ Signature change`: `List` → JSON (was comma-joined) and `null` → omitted (was empty-string part) for `resumeExecution` / `validate*` — no signature change, but a behavior change for existing callers.

**Go.** Corrected the scope note to say plainly that Go is impacted in *serialization* (the `default: fmt.Sprint(v)` in `api_executions.go` produces `map[count:2]`), not just that the `.FormData` surface exists — same defect, sized as a follow-up since this PR is Java-only per the issue.

Local: main + test compile clean against `local-ci-version`; leaving the live suite to run in CI.

Issue #434, opened 2026-09-16
Go SDK: flows CSV export requests text/plain (should be text/csv) + distinct-field-values test enum casing


jymaire:
Surfaced while reviewing the SDK-coverage work under EPIC #331 (Go/Java SDKs, already merged via #420/#429 — outside the Python PR #433). Each item below was verified against the backend controllers in `kestra-io/kestra` / `kestra-io/kestra-ee`.

## 1. Go `ExportFlowsByQueryCsv` sends `Accept: text/plain` instead of `text/csv` (confirmed)

`go-sdk/kestra_api_client/flows_api.go:413`
```go
return a.doText(ctx, "GET", path, params)   // no accept override → doText defaults to contentPlainText
```
`doText` (`base_api.go:365-366`) defaults `accept` to `contentPlainText`. The sibling exporter passes the CSV type explicitly:

`go-sdk/kestra_api_client/executions_api.go:850`
```go
return a.doText(ctx, "GET", path, params, contentCSV)
```
Both backend routes are declared `produces = MediaType.TEXT_CSV`:
- `kestra` `ExecutionController.java:2767` — `@Get(uri = "/export/by-query/csv", produces = MediaType.TEXT_CSV)`
- `kestra` `FlowController.java:978` — `@Get(uri = "/export/by-query/csv", produces = MediaType.TEXT_CSV)`

So `ExportFlowsByQueryCsv` requests the wrong media type; against content negotiation this risks `406 Not Acceptable`, and it's inconsistent with the executions sibling regardless.

**Fix:** pass `contentCSV` to `doText`, matching `ExportExecutionsByQueryToCsv`. (The Python equivalent `export_flows_by_query_csv` already sends `accept=self.CSV` correctly.)

## 2. Go test hardcodes uppercase `"NAMESPACE"` for the distinct-field-values field (minor / verify)

`go-sdk/test/api_executions_test.go:1200`
```go
values, err := KestraTestClient().Executions().FindDistinctExecutionFieldValues(ctx, MAIN_TENANT, "NAMESPACE", nil, kestra_api_client.PtrInt(500))
```
The SDK's own constant is lowercase — `QUERYFILTERFIELD_NAMESPACE QueryFilterField = "namespace"` (`model_query_filter_field.go:31`) — and the backend binds `@QueryValue QueryFilter.Field field` (`kestra` `ExecutionController.java:340`). The live-validated Python test calls the same endpoint with the lowercase wire value and passes.

[...]

PR #447, opened 2026-09-18
fix(go): flows CSV export requests text/csv + custom blueprint source requests yaml (#434)

jymaire:
## What

Fixes the two Go SDK items in #434, plus one adjacent same-class fix found during the required call-site sweep.

### 1. `ExportFlowsByQueryCsv` requested `text/plain` instead of `text/csv`
`flows_api.go` called `doText(ctx, "GET", path, params)` with no accept override, so `doText` fell back to its `contentPlainText` default. The backend route is declared `produces = MediaType.TEXT_CSV`, so this risked a `406 Not Acceptable` under content negotiation and was inconsistent with the executions sibling `ExportExecutionsByQueryToCsv`, which passes `contentCSV`. Now passes `contentCSV`.

### 2. Distinct-field-values test used a hardcoded uppercase `"NAMESPACE"`
`api_executions_test.go` passed the literal `"NAMESPACE"`, but the SDK constant and wire value are lowercase (`QUERYFILTERFIELD_NAMESPACE = "namespace"`) and the backend binds `@QueryValue QueryFilter.Field`. Switched to `string(kestra_api_client.QUERYFILTERFIELD_NAMESPACE)` — correct regardless of whether Micronaut currently binds case-insensitively, and guards against a latent break if binding tightens. The SDK method signature (`field string`) is unchanged.

### 3. (Sweep finding) `InternalBlueprintFlow` custom-blueprint source requested `text/plain` instead of `application/yaml`
Task 3 of the approved plan required sweeping the remaining `doText` call sites that rely on the `text/plain` default. `InternalBlueprintFlow` (custom blueprint source) defaulted to `text/plain`, while its own community sibling `BlueprintSource` sends `application/yaml` and the Python SDK's `internal_blueprint_flow_source` sends `accept=self.YAML`. Fixed to `application/yaml`.

### Sweep result (all `doText` call sites relying on the default)
| Call site | Accept | Verdict |
|---|---|---|
| `blueprints_api.go:20` community source | explicit `application/yaml` | ✅ correct |
| `blueprints_api.go:74` custom source | was default `text/plain` | 🔧 **fixed → `application/yaml`** |
| `flows_api.go` CSV export | was default `text/plain` | 🔧 **fixed → `text/csv`** |
| `plugins_api.go:46` icon.svg | explicit `image/svg+xml` | ✅ correct |
| `plugins_api.go:96` `pluginUi/{path}` | default `text/plain` | ✅ left as-is — polymorphic UI asset (Java returns opaque `byte[]`), no single correct media type |

## Verification
- `go build ./...` — success
- `go vet ./...` — no issues

> ⚠️ Live Go tests were **not** run here (require a Kestra instance; run by the maintainer). Note per the plan: `exportFlowsByQueryCsvTest` **skips on 403** because flow CSV export is permission-gated on the EE test image, so a green run does not prove the Accept fix — it must be verified by inspecting the outgoing request header.

## ⚠️ Signature change
None. `ExportFlowsByQueryCsv`, `InternalBlueprintFlow`, and `FindDistinctExecutionFieldValues` signatures are all unchanged.

## Out of scope
- Consolidating the `doCSV` helper with the `doText(..., contentCSV)` idiom (and migrating `triggers_api.go:31`).
- Widening `FindDistinctExecutionFieldValues`'s `field string` param to the typed `QueryFilterField` (would be breaking).
- The upstream export-permission gap causing the flows CSV test to skip.
- Java's `export` CSV accept type (not audited here).

part-of: https://github.com/kestra-io/client-sdk/issues/434

---
*[View as Artifact](https://claude.ai/artifact/HHt6PoD6qwcGTqS6JQUYuG)*


jymaire:
## Code review — resolution

Ran `/code-review` on the diff. Three findings; disposition below.

**1. Distinct-field-values test now sends `field=namespace` (lowercase) — possible bind regression.** Kept the change (it's what the issue requested and matches the SDK's own `QUERYFILTERFIELD_NAMESPACE = "namespace"`). Micronaut's `CharSequence→Enum` converter does an exact `Enum.valueOf` then a case-insensitive name fallback, so `"namespace"` binds to the `NAMESPACE` constant the same as `"NAMESPACE"` did. ⚠️ **Caveat for the reviewer:** this is the one runtime-behavior change and I couldn't confirm it against a live 2.0 backend here — worth a glance at the next CI run of `findDistinctExecutionFieldValuesTest`. (Note: the issue's claim that "the Python live test passes with lowercase" is unsubstantiated — Python has no `find_distinct_field_values` implementation, only generator doc boilerplate.)

**2. `"application/yaml"` literal vs the `contentYAML = "application/x-yaml"` constant.** Not a defect. `contentYAML` is for YAML request *bodies* (flow uploads); the blueprint *source* endpoints send `application/yaml`, and line 74 now matches its already-working community sibling at line 20 exactly. Using `contentYAML` here would diverge from the working sibling.

**3. No regression test pinning the Accept header.** Fixed in c7b55601 — added `TestTextEndpointAcceptHeaders`, a mock-server test asserting the exact Accept header for both endpoints. This matters because `exportFlowsByQueryCsvTest` self-skips on 403 (export permission-gated on the EE image) and never exercises the header. Verified it fails (`text/csv` → `text/plain`) when the `doText` override is dropped.

`go build`, `go vet`, and the new test all pass.

Malaydewangan09 (approved):

PR #452, opened 2026-09-21
build(javascript): bump @kestra-io/hey-api-plugin to v0.3.0

### LINKED ISSUE #444: JS SDK: application/yaml migration will regress #340 on next regeneration
## Context

With the Micronaut migration on the backend, all YAML endpoints move from the non-standard `application/x-yaml` media type to the RFC-compliant `application/yaml`. The old `application/x-yaml` **remains supported** server-side, and the change is only reflected in the OpenAPI spec (which will now advertise `application/yaml`).

**Runtime impact: none.** The JS SDK currently sends `application/x-yaml` for all YAML-source bodies, and the server still accepts it. This issue is about the **next JavaScript regeneration** against the migrated spec.

## Cause

The JS SDK is the only generated SDK. To work around hey-api preferring the `application/json` request-body variant over YAML (issue #340), we run the `fixYamlSourceRequestBodyContentType` patch from `@kestra-io/hey-api-plugin`. That patch keys off the **literal** media-type string:

```js
// @kestra-io/hey-api-plugin — dist/index.mjs
const YAML_MEDIA_TYPE = "application/x-yaml";

function fixYamlSourceRequestBodyContentType(method, path, operation) {
  const content = operation?.requestBody?.content;
  if (!content || typeof content !== "object") return;
  if (!isPlainString(content[YAML_MEDIA_TYPE]?.schema)) return;   // ← guard fails when spec says application/yaml
  if (!isPlainString(content[JSON_MEDIA_TYPE]?.schema)) return;
  delete content[JSON_MEDIA_TYPE];                                 // drops the JSON variant
  // ...reorder so YAML is the resolved body...
}
```

Once the EE OpenAPI spec flips YAML request bodies from `application/x-yaml` to `application/yaml`:

1. `content["application/x-yaml"]` is `undefined`, so the guard returns early.
2. The patch no longer drops the `application/json` variant nor forces the YAML content type.
3. hey-api falls back to preferring `application/json` — i.e. **the exact #340 bug regresses**: YAML source bodies get sent as JSON, breaking flow/policy/app/etc. create & update.

Also affected (secondary):
- `sdk-freshness-check` will flag drift across every YAML endpoint (bundled spec says `x-yaml`, live EE says `yaml`) until the bundled `kestra-ee.yml` is refreshed.

## Fix

Update the shared `@kestra-io/hey-api-plugin` so `fixYamlSourceRequestBodyContentType` recognizes **both** media types and normalizes to one:

- Match `application/yaml` **and** `application/x-yaml` when detecting the YAML request body.
- Normalize the emitted `Content-Type` to a single value. Since the server accepts both, we can either:
  - keep sending `application/x-yaml` (smallest diff, matches Java/Python/Go), or
  - move to `application/yaml` to align with the RFC-compliant spec (preferred long-term; then align the other SDKs' `YAML` constants too).

Sketch:

```js
const YAML_MEDIA_TYPES = ["application/yaml", "application/x-yaml"];
const CANONICAL_YAML = "application/x-yaml"; // or "application/yaml"

function fixYamlSourceRequestBodyContentType(method, path, operation) {
  const content = operation?.requestBody?.content;
  if (!content || typeof content !== "object") return;
  const yamlType = YAML_MEDIA_TYPES.find(t => isPlainString(content[t]?.schema));
  if (!yamlType) return;
  if (!isPlainString(content[JSON_MEDIA_TYPE]?.schema)) return;
  const yamlSchema = content[yamlType];
  delete content[JSON_MEDIA_TYPE];
  delete content[yamlType];
  const reordered = { [CANONICAL_YAML]: yamlSchema };
  for (const [mt, v] of Object.entries(content)) if (mt !== CANONICAL_YAML) reordered[mt] = v;
  operation.requestBody.content = reordered;
}
```

## Acceptance criteria

- [ ] `fixYamlSourceRequestBodyContentType` handles a spec that advertises `application/yaml` (and still handles `application/x-yaml`).
- [ ] Regenerating JS against the migrated spec keeps YAML-source bodies (flows, policies, apps, dashboards, namespaces, reusable inputs, test suites) sending a YAML content type — not `application/json`.
- [ ] Refresh bundled `kestra-ee.yml` and confirm `sdk-freshness-check` is green.
- [ ] JS vitest for the YAML create/update endpoints passes (they assert real values, not just `toBeDefined`).

## Notes

- Hand-written SDKs (Java/Python/Go) send `application/x-yaml` via hardcoded constants and are **not** at runtime risk (old type still accepted). Aligning them to `application/yaml` is optional follow-up.
- Related: #340 (original YAML-vs-JSON request-body fix this guards against regressing).

elevatebart:

Bumps `@kestra-io/hey-api-plugin` from v0.1.0 to v0.3.0 in `javascript/javascript-sdk/package.json` and the lockfile. v0.3.0 carries the YAML media-type fix from https://github.com/kestra-io/kestra/pull/19632: `fixYamlSourceRequestBodyContentType` now matches `application/yaml` as well as `application/x-yaml` and normalizes to the latter.

Without it, the next JS regeneration against a Micronaut-migrated spec falls back to the `application/json` request-body variant and re-opens #340, sending YAML source bodies with a JSON content type. Closes #444.

`openapi-ts.config.ts` needs no change: it already calls `fixYamlSourceRequestBodyContentType`, so the fix arrives through the dependency. If #445 merges after this, its local `normalizeYamlRequestBodyMediaType` is dead code and should be dropped.

### ⚠️ Signature change

The 0.1.0 → 0.3.0 jump also picks up v0.2.0's runtime change, which is bundled into this SDK's `dist`:

`createConfigureClient(client, formDataBodySerializer)` → `createConfigureClient(client, formDataBodySerializer, enterpriseFeature?)`

plus a new `EnterpriseFeatureError` export. The parameter is optional and `src/index.ts:145` passes two arguments, so behavior is unchanged. Wiring an EE-feature registry into the client (so a 404 on an EE-only route says so instead of raising a generic error) is a separate decision, not taken here.

---

The first commit 404'd in CI because v0.3.0 had no release tarball yet. kestra-io/kestra#19632 is now merged and the *Release hey-api-plugin* workflow published it, so the second commit pins the real `integrity` (sha512 of the published archive, cross-checked against what `npm install` resolves on its own).

Verified locally against the published tarball:

- `npm ci` installs v0.3.0 and the full `./generate-sdks.sh javascript` run completes.
- `src/openapi/` regenerated with v0.3.0 is **byte-identical** to a regeneration with v0.1.0. The bump changes no SDK output at all.
- 30 endpoints still emit `Content-Type: application/x-yaml` for YAML-source bodies, so #340 stays fixed.

The plugin-side behavior is covered upstream by `ui/packages/hey-api-plugin/tests/patch.spec.ts` (8 cases: legacy `x-yaml` + JSON, migrated `yaml` + JSON, both spellings, YAML-only, JSON-only, a `format: binary` schema, schema preservation, no request body).

The coverage bot's 90.3% report on this PR is the repo's standing baseline, untouched by a lockfile change.

https://claude.ai/code/session_0159TK4t9hREmz5TtiruRJ5k


elevatebart:
`Test Javascript SDK` fails, as expected, and no push to this branch can fix it.

`./generate-sdks.sh javascript` gets through the spec customizer fine and dies on `npm ci`:

```
npm error code E404
npm error 404 Not Found - GET https://github.com/kestra-io/hey-api-plugin/releases/download/v0.3.0/kestra-io-hey-api-plugin-0.3.0.tgz
```

v0.3.0 has no release tarball yet. It appears once the **Release hey-api-plugin** workflow is dispatched on kestra-io/kestra#19632's branch (that PR bumps the package to 0.3.0 and is green, waiting on review). Not re-running the job: it would 404 identically.

Two things follow, in order, and then this goes green on its own:

1. Merge or approve kestra-io/kestra#19632 and dispatch the release workflow.
2. `npm install` here to write the lock entry's `integrity`, which is deliberately absent rather than guessed, then push.

I have not dispatched the release myself: it would publish a tarball from an unmerged branch.

---

jymaire (approved):

Issue #246, opened 2026-05-28
Add complex query filter support (AND/OR + nested groups)


ammeek:
# Description

Kestra core PR [#16197](https://github.com/kestra-io/kestra/pull/16197) introduced **complex condition-group support** for the `filters[...]` query parameters used by every `*ByQuery` / search endpoint.

The backend now accepts:

- `AND` / `OR` groups
- one level of nesting
- fully backward-compatible flat filters

The client SDKs currently expose only the flat filter shape (a list of single `field/op/value` filters). This issue tracks adding **first-class support for grouped and nested queries** in:

- Python
- Java
- Go
- JavaScript

The goal is to allow SDK users to construct the same queries the UI now produces **without manually encoding query strings**.

---

# API Query Shape

As introduced in [#16197](https://github.com/kestra-io/kestra/pull/16197), repeated `filters[...]` keys encode boolean structure through bracketed segments.

## Flat Filters (Legacy)

This format must continue to work unchanged.

```
filters[flowId][EQUALS]=flow-id
```

## Top-Level `AND` Group

```
filters[and][0][namespace][EQUALS]=ns
filters[and][1][flowId][EQUALS]=f
```

## Top-Level `OR` Group

```
filters[or][0][scope][EQUALS]=s1
filters[or][1][scope][EQUALS]=s2
```

## Nested Groups (One Level)

Example: `AND` containing an `OR`.

```
filters[and][0][namespace][EQUALS]=ns
filters[and][1][or][0][scope][EQUALS]=s1
filters[and][1][or][1][scope][EQUALS]=s2
```

---

# Serialization Rules

Each leaf node follows this structure:

```
filters[<fieldName>][<OP>]=value
```

Where:

- `<fieldName>` uses `camelCase`
- `<OP>` is uppercase

## Special Cases

| Logical Field | Serialized Name |
| --- | --- |
| `QUERY` | `q` |
| `MIN_LEVEL` | `level` |

## Value Encoding

| Operator / Type | Encoding |
| --- | --- |
| `IN` / `NOT_IN` | CSV values |
| `Map` values | `[key]` suffixes |
| `LABELS` | Uses map-style serialization |

---

# Proposed User-Facing API

A jOOQ-style DSL has already been prototyped for Java on the local `feat/complex-query-java` branch.

Other SDKs should implement idiomatic equivalents.

## Java Prototype

```java
import static io.kestra.sdk.query.Query.*;

List<QueryFilter> filters = where(
    and(
[...]

PR #453, opened 2026-09-21
feat: complex query filter support (AND/OR + nested groups) for Go, Java, Python (#246)

jymaire:
## What

Adds first-class **AND / OR + one-level-nested query filter** support to the **Go, Java and Python**
SDKs, consuming the bracketed `filters[...]` wire format introduced by
[kestra#16197](https://github.com/kestra-io/kestra/pull/16197).

Users can now build the same grouped queries the UI produces, without hand-encoding query strings:

```java
// Java
List<QueryFilter> filters = where(and(
    eq(QueryFilterField.NAMESPACE, "company.team"),
    or(eq(QueryFilterField.STATE, "SUCCESS"), eq(QueryFilterField.STATE, "WARNING"))));
client.executions().searchExecutions(/* ..., */ filters);
```

## How

Each SDK gets three additive pieces (no parallel AST type — the spec's own `QueryFilter` schema is
the AST once `logical`/`children` are added):

1. **Model** — additive nullable `logical` (`QueryFilterLogical` = `and`/`or`) + recursive `children`
   on `QueryFilter`. Python already carried both. Body-based `*ByQuery` endpoints get grouped filters
   for free via plain JSON.
2. **Recursive serializer** — the flat leaf loop becomes a walk with a bracket-path prefix.
3. **`query` DSL** — `where/and/or/filter` + `eq, notEq, in, notIn, contains, startsWith, endsWith,
   regex, prefix, greaterThan/greaterThanOrEqual/lessThan/lessThanOrEqual` (idiomatic per language),
   returning the SDK's existing filter list type.

### Wire-format contract (the important part)

The serialization was reverse-engineered from the **UI encoder** in kestra#16197
(`.../filter/utils/helpers.ts` `encodeFilterGroupsToQuery`), which is the ground truth the backend
parses. The key rule, pinned by shared golden vectors:

- A **top-level AND whose children are all leaves flattens to the bare `filters[field][OP]` form** —
  byte-identical to the legacy flat filter list. This is the core backward-compat invariant.
- Explicit `filters[and][i]` / `filters[or][i]` appears **only** when the top logical is OR or a
  nested group is present, e.g.:
  ```
  filters[and][0][namespace][EQUALS]=ns
  filters[and][1][or][0][scope][EQUALS]=s1
  filters[and][1][or][1][scope][EQUALS]=s2
  ```
- A single-child group flattens to that child; an empty group emits nothing; nesting deeper than one
  level raises a clear client-side error; a node that is both a leaf and a group is rejected.

> Note: the issue's "Top-Level AND Group" example shows the explicit `filters[and][0]` form for
> all-leaf AND, but the backend accepts both and the UI emits the flat form; we emit flat to keep
> existing queries wire-identical. Documented in the golden fixture.

### Shared golden vectors

`test-utils/query-filter-golden.json` lists `(query tree → expected ordered key=value pairs)` for
flat, top-level AND, top-level OR, AND-containing-OR, OR-containing-AND, single-child, empty group,
IN CSV (flat + grouped), LABELS map inside a group, and QUERY→`q`. **All three SDKs assert against
the same file** so the implementations cannot drift.

### Field-name mapping (resolved open question from the plan)

Confirmed that field→wire mapping stays each SDK's existing concern and needs no new code: Java
`filterFieldName` maps `QUERY→q`/`MIN_LEVEL→level`; Python's enum values are already wire-form
(`q`, `level`); Go bakes wire values into `SearchFilterField`. The recursive serializer reuses each
SDK's existing leaf emission unchanged.

## Tests

Offline serializer unit tests per SDK, driven by the shared golden vectors (no running Kestra):

| SDK | command | result |
|---|---|---|
| Go | `cd go-sdk && go build ./... && go test ./... -run QueryFilter` | 18 passed |
| Python | `cd python/python-sdk && python -m pytest tests/test_query_filter_golden.py -q` | 16 passed |
| Java | `cd java/java-sdk && ./gradlew cleanTest test --tests '*QueryFilterGolden*'` | 15 passed |

Each suite also covers `MIN_LEVEL→level`, over-deep nesting → error, and the ambiguous leaf+group
node → error. Go additionally verifies duplicate-key survival (`Add` not `Set`) and map-key sort
determinism.

## ⚠️ Signature change

Per `AGENTS.md`, diffed every touched public method:

- **No public `*ByQuery` / search method signature changed** — every change is additive (`where(...)`
  returns the existing `List<QueryFilter>` / `list[QueryFilter]` / `[]SearchFilter`).
- **No existing `QueryFilter` constructor changed** — `logical`/`children` are additive nullable
  fields.
- Internal serializer helpers changed (not part of any SDK's public API contract):
  - Python: `append_filter_params(params, filters)` → `append_filter_params(params, filters, prefix="filters")` (defaulted, backward-compatible).
  - Java: `ApiClient#parameterToPairs` signature unchanged; leaf emission extracted into private helpers.
- **Naming note (Go):** the generic DSL constructor is exported as **`FilterBy`**, not `Filter`,
  because `Filter` is already a generated model type in the package. All other helper names match
  across the three SDKs.

### ⚠️ Signature changes from the review follow-up (`f939b6e`)

These are deliberate and touch only **new-in-this-PR** surface (no released consumer affected):

- **Go — `AppendFilterParams(params, filters)` now returns `error`** (was `func(...)` returning nothing and panicking on a malformed tree). `Where(...)` no longer panics at construction either. All internal callers — ~70 hand-written search methods and 25 generated `*ByQuery` Execute methods — propagate it through their existing `error` return, so no public *method* signature changed; only this serializer helper. The DSL constructors (`Eq`/`And`/`Or`/`Where`/…) keep returning `[]SearchFilter` and stay composable.
- **Go — legacy filter serialization switched from `params.Set` to `params.Add`** (via the new ordered emitter): two flat filters on the same field+op now **both** reach the wire instead of the last one winning. Correct and required for OR, but a behavior change worth calling out.
- **Python — `kestrapy` package root no longer re-exports the query-filter DSL helpers** (`eq`, `not_eq`, `in_`, `contains`, `regex`, `prefix`, `gt`, `lt`, `where`, `and_`, `or_`, …). Import them from `kestrapy.query` instead. This removes the `regex` root-name collision with the `regex` package the models import. (DSL is new in this PR, so nothing released depended on the root names.)
- **Go — `ParseQueryFilters` is now deprecated** (kept exported, still `(map[string]string, error)`); prefer the internal ordered path. It no longer nil-derefs on group nodes but still collapses same-key filters due to its map return — noted in its doc comment.

## 🧪 Live QA sign-off

Verified against a real **kestra-ee v2.0.2** instance (released after kestra#16197 merged, so it
carries the feature). Seeded 3 flows: `qa.alpha.orders` (`tier: gold`), `qa.alpha.payments`
(`tier: silver`), `qa.beta.orders` (`tier: gold`).

**Query-param path** — real returned IDs (identical results via raw `curl` of the exact wire strings
the SDKs emit **and** via a real Python-SDK `flows.search_flows(..., filters=...)` call built with the
`query` DSL):

| Query | Emitted wire | Returned flows |
|---|---|---|
| baseline (flat/legacy) | `filters[namespace][EQUALS]=qa.alpha` | `qa.alpha.orders`, `qa.alpha.payments` |
| **OR** (union) | `filters[or][0][namespace][EQUALS]=qa.alpha` `filters[or][1][namespace][EQUALS]=qa.beta` | `qa.alpha.orders`, `qa.alpha.payments`, `qa.beta.orders` |
| **AND(OR)** (intersection) | `filters[and][0][namespace][EQUALS]=qa.alpha` `filters[and][1][or][0][labels][EQUALS][tier]=gold` `filters[and][1][or][1][labels][EQUALS][tier]=silver` | `qa.alpha.orders`, `qa.alpha.payments` |
| control: labels OR **without** the ns AND | `filters[or][0][labels][EQUALS][tier]=gold` `filters[or][1][labels][EQUALS][tier]=silver` | `qa.alpha.orders`, `qa.alpha.payments`, **`qa.beta.orders`** |

The control returns `qa.beta.orders` but AND(OR) does not — proving the AND genuinely intersects the
nested OR rather than the server ignoring it. `[labels][EQUALS][tier]` map serialization and the
legacy flat form both confirmed. All HTTP 200.

**JSON-body path** — a real `triggers.delete_triggers_by_query` call with a grouped body was accepted
by the server (`{"totalItems":0}`, no 4xx), confirming the model's `logical`/`children` marshalling
round-trips:
```json
{"filters":[{"field":"namespace","operation":"EQUALS","value":"qa.alpha"},
 {"logical":"or","children":[{"field":"labels","operation":"EQUALS","value":{"tier":"gold"}},
                             {"field":"labels","operation":"EQUALS","value":{"tier":"silver"}}]}]}
```

> QA was driven from Python (which shares the exact wire output pinned by the golden vectors, so it
> validates the Go/Java serializers' output too); the raw-`curl` column is language-agnostic proof the
> backend accepts that exact wire. Instance was a throwaway v2.0.2 container, torn down after.

## Scope

This PR covers **Go + Java + Python**. **JavaScript is intentionally excluded**: its serializer lives
in the external `@kestra-io/hey-api-plugin` package and must be changed + released there first, then
pinned here (tracked in the issue). No `javascript/` files are touched.

part-of: https://github.com/kestra-io/client-sdk/issues/246


jymaire:
### Pre-review status

- **Offline golden vectors green** across all three SDKs (Go 21 · Java 17 · Python 18 tests), driven by the shared `test-utils/query-filter-golden.json`.
- **Live QA done** against a throwaway **kestra-ee v2.0.2** instance (carries kestra#16197): OR→union, AND(OR)→intersection with real returned IDs, plus the JSON-body path accepted — see the *Live QA sign-off* section in the description.
- **Self code-review run**; the follow-up commit addresses 5 cross-SDK divergences it found (chiefly: Go was silently emitting an unfiltered query on a malformed filter tree where Java/Python raise — now panics loudly; any-map-type handling; raw-struct normalization; null value/field consistency).

JS is intentionally out of scope (lives in `@kestra-io/hey-api-plugin`).

go-sdk/kestra_api_client/model_query_filter.go
```diff
@@ -16,9 +16,11 @@ var _ MappedNullable = &QueryFilter{}
...
 // QueryFilter struct for QueryFilter
 type QueryFilter struct {
-	Field                *QueryFilterField `json:"field,omitempty"`
-	Operation            *QueryFilterOp    `json:"operation,omitempty"`
-	Value                any               `json:"value,omitempty"`
+	Field                *QueryFilterField   `json:"field,omitempty"`
+	Operation            *QueryFilterOp      `json:"operation,omitempty"`
+	Value                any                 `json:"value,omitempty"`
+	Logical              *QueryFilterLogical `json:"logical,omitempty"`
+	Children             []QueryFilter       `json:"children,omitempty"`
```
Malaydewangan09, line 23: Adding `Logical`/`Children` here never reaches the wire, and makes things worse.

The 25 generated `*ByQuery` endpoints (`api_executions.go`, `api_flows.go`, `api_triggers.go`, `api_kv.go`, `api_namespaces.go`) serialize `[]QueryFilter` through `ParseQueryFilters` in `legacy_client.go:246`, which this PR doesn't touch. It does `*qf.Field` unconditionally, so a group node nil-derefs:

```
PROBE group-through-ParseQueryFilters: PANIC runtime error: invalid memory address or nil pointer dereference
PROBE dup-leaves: map[filters[namespace][EQUALS]:b]   <- second filter ate the first
```

The new recursive serializer only covers `[]SearchFilter`, which these endpoints never use. So "body-based `*ByQuery` endpoints get grouped filters for free" isn't true for Go, they are query-param serialized through the old path. Same function returns `map[string]string`, so same-key filters collapse and emission order is random.

java/java-sdk/src/main/java/io/kestra/sdk/internal/ApiClient.java
```diff
@@ -699,7 +681,189 @@ public List<Pair> parameterToPairs(String collectionFormat, String name, Collect
...
+      for (QueryFilter leaf : units) {
+        params.addAll(emitLeaf("filters", leaf, escape));
+      }
+      return params;
+    }
+
+    // Grouped form: filters[<logical>][i]...
+    for (int i = 0; i < units.size(); i++) {
+      QueryFilter unit = units.get(i);
+      String unitPrefix = "filters[" + topLogical.getValue() + "][" + i + "]";
```
Malaydewangan09, line 754: NPE when a group has `children` but no `logical`. `topLogical` comes straight from `group.getLogical()` on line 727 with no null guard:

```
Cannot invoke "io.kestra.sdk.model.QueryFilterLogical.getValue()" because "topLogical" is null
```

Reachable from the public builders this PR adds (`addChildrenItem`, `children(...)`) and from JSON deserialization. Python treats the same input as an implicit AND and serializes it fine.

Either default to AND or throw `ApiException(400)` like every other structural error in this method. An unchecked NPE is the one outcome the surrounding code set out to avoid. Line 762 needs the same guard for `unit.getLogical()`.

java/java-sdk/build.gradle
```diff
@@ -118,13 +118,21 @@ dependencies {
     implementation "org.openapitools:jackson-databind-nullable:$jackson_databind_nullable_version"
     implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:$jackson_version"
     testImplementation "junit:junit:$junit_version"
+    testImplementation 'org.junit.jupiter:junit-jupiter:5.13.4'
+    // Gradle 9 no longer puts the JUnit Platform launcher on the test runtime
+    // classpath implicitly; without it the test worker fails to start.
+    testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
```
Malaydewangan09, line 129: No version, and nothing supplies one. This resolves in CI only because CI's `mavenLocal()` is empty, so Gradle fetches `junit-jupiter-5.13.4.module` from Central and picks up the BOM constraints.

On any machine that already has the jupiter POM cached in `~/.m2` (and `mavenLocal()` is first at line 20) `./gradlew check` fails outright:

```
> Could not find org.junit.platform:junit-platform-launcher:.
```

Hit this locally on a clean checkout. The BOM fixes it and the 17 tests run:

```groovy
testImplementation platform('org.junit:junit-bom:5.13.4')
testImplementation 'org.junit.jupiter:junit-jupiter'
testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
```

python/python-sdk/kestrapy/query_filter.py
```diff
@@ -27,63 +27,209 @@ def _to_camel_case(s: str) -> str:
...
+def _logical_str(f: Any) -> str:
+    """Wire form of a group's logical operator: lowercase 'and'/'or'."""
+    lg = _enum_value(_node_logical(f))
+    return str(lg).lower() if lg is not None else 'and'
+
+
+def _classify(f: Any) -> str:
+    """Return 'leaf' or 'group'; raise if a node is ambiguously both."""
+    has_field = _node_field(f) is not None
+    has_group = _node_logical(f) is not None or _node_children(f) is not None
```
Malaydewangan09: The three SDKs use three different definitions of "is this a group", so malformed nodes behave differently. Go checks `Logical != nil`, Java checks `logical != null || !children.isEmpty()`, Python checks `logical is not None or children is not None`.

| input | Java | Python | Go |
|---|---|---|---|
| leaf + `children=[]` | leaf, ok | ValueError | leaf, ok |
| `children=[a,b]`, no `logical` | NPE | AND group, ok | panic "requires a field" |

All three probed, not inferred. The golden vectors pin the happy path but nothing pins the error path, which is where the drift the fixture exists to prevent actually shows up. Worth picking one rule and adding these two cases to the fixture.

README_GO_SDK.md
```diff
@@ -96,3 +96,31 @@ would never be resolved by `go get` — so the per-language suffix convention
...
+)
+// filters is a []SearchFilter — pass it to the search / *ByQuery calls that accept filters.
+```
+
+- Helpers: `Where`, `And`, `Or`, `Eq`, `NotEq`, `In`, `NotIn`, `Contains`, `StartsWith`,
+  `EndsWith`, `Regex`, `Prefix`, `Gt`, `Gte`, `Lt`, `Lte`, and the generic `FilterBy(field, op, value)`
+  (named `FilterBy` rather than `Filter` because `Filter` is an existing model type).
+- **Backward compatible:** a plain `[]SearchFilter` (or `Where(And(...leaves))`) serializes to the
+  same flat `filters[field][OP]=value` wire format as before.
+- Nesting is **one level deep** (an `And` containing an `Or`, or vice-versa); deeper nesting returns
```
Malaydewangan09: It panics, it doesn't return an error. `Where()` panics at construction and `AppendFilterParams` panics at call time.

Since every search method already returns `error`, plumbing the error out would be better than fixing the doc. A Go library panicking on input-derived filters is a crash in the caller's request handler. There is precedent at `legacy_client.go:290`, but this widens the surface a lot now that it fires from the DSL.

Malaydewangan09 (commented):
Went through this end to end: built and ran all three suites, probed the edge cases, and checked the wire format against the UI encoder (`KsDataTable/filter/utils/helpers.ts`) and the backend binder (`QueryFilterFormatBinder`) in kestra core.

The design is right. Shared golden fixture is the correct mechanism, the flat-AND backward-compat invariant genuinely matches the UI encoder's `onlyOneLeaf` branch, and the live QA control query (labels OR without the namespace AND) is a real test rather than a screenshot of a 200. Suites I ran: Go 23 pass, Python 19 pass, Java 17 pass (after fixing the build, see the `build.gradle` comment).

Three things block, inline below. Plus a few smaller ones:

**Python has the same second-serializer problem as Go.** `kestrapy/api_client.py:495-531` and `:570-602` carry a flat-only copy that raises `ApiValueError('Filter elements must contain field and operation')` on a group node. `ApiClient` is unused internally but exported from `kestrapy/__init__.py:707`, so it is public surface. Update it or mark it unsupported.

**Nits:**
- `kestrapy/__init__.py` now exports `eq`, `contains`, `regex`, `prefix`, `in_`, `gt`, `lt` at the package root. `regex` collides with the `regex` package the SDK's own models import. `from kestrapy.query import ...` would keep the root clean.
- The Java golden test only asserts `collectFilterPairs(..., escape=false)`. The escaped path is what production uses and is untested for the new group prefixes.
- Nothing tests the DSL constructors' own drop/flatten rules. Java's `isEmptyGroup` (`field == null && children empty`) and Python's `_is_empty_group` (`logical is not None and not children`) are different predicates and nothing catches that.
- Go's DSL only builds `[]SearchFilter`, so the 25 `[]QueryFilter` endpoints get no DSL. Java and Python use one type for both.
- Backend default is `maxDepth 3` / `maxWidth 20` (`QueryFilterConfiguration`), the UI caps at `MAX_RENDERABLE_NESTING_DEPTH = 2`. The SDK cap of one nested level is a fine scope call, but the READMEs read as if the server refuses deeper. One sentence would clear that up.
- `params.Set` to `params.Add` in Go changes legacy behaviour: two flat filters on the same field+op now both reach the wire instead of the last one winning. Correct, and needed for OR, but it belongs in the signature-change section.

Malaydewangan09 (approved):

go-sdk/kestra_api_client/model_query_filter.go
jymaire: Fixed in f939b6e — the QueryFilter/`*ByQuery` path is now a first-class citizen instead of a flat-only dead end.

- New `queryFilterToSearchFilter` converter + `buildQueryFilterParams` in `legacy_client.go` route `[]QueryFilter` through the **same** recursive serializer as the SearchFilter path (full AND/OR + one-level nesting), so grouped filters now actually reach the wire.
- All 25 generated `*ByQuery` Execute methods now call `addFilterQueryParams(localVarQueryParams, *r.filters)`, which appends via `url.Values.Add` — no more `map[string]string` collapse (same-field OR duplicates survive, emission order is stable) and no more `*qf.Field` nil-deref on a group node.
- The reflection `filters` branch in `parameterAddToHeaderOrQuery` (which panicked) is gone; structural errors surface through the method's existing `error` return.
- `ParseQueryFilters` stays exported (deprecated) and no longer nil-derefs, but keeps its `map[string]string` shape so it still collapses same-key filters — flagged in its doc comment; the real path preserves them.

java/java-sdk/src/main/java/io/kestra/sdk/internal/ApiClient.java
jymaire: Fixed in f939b6e. `collectFilterPairs` no longer reads `getLogical().getValue()` unguarded: a new `logicalOrAnd(QueryFilter)` helper defaults a null logical to `AND` at both the top-group site and the nested-group site (line ~762). This matches Python's implicit-AND behavior — see the cross-SDK unification below. Reachable via the public builders and JSON deserialization no longer NPEs.

java/java-sdk/build.gradle
jymaire: Fixed in f939b6e. Switched to the JUnit BOM so the launcher version resolves regardless of `~/.m2` state:
```groovy
testImplementation platform('org.junit:junit-bom:5.13.4')
testImplementation 'org.junit.jupiter:junit-jupiter'
testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
```
Thanks for catching the cached-POM failure.

python/python-sdk/kestrapy/query_filter.py
jymaire: Fixed in f939b6e. Picked one rule and pinned it in the shared fixture.

**Unified rule (all 3 SDKs):** a node is a **group** iff `logical != null` **OR** it has a **non-empty** children list. A group with a null logical defaults to **AND**. An empty children list on a leaf is ignored (stays a leaf). A node carrying both a field and group data is ambiguous → client-side error.

- Go: `isGroup`/`isAmbiguous` now use `logical != nil || len(children) > 0`; nil logical defaults to AND in `buildFilterParams`.
- Java: `isGroupNode` unchanged (already `logical != null || children non-empty`); the NPE from a null logical is fixed via `logicalOrAnd` (defaults AND).
- Python: `_classify` now uses `bool(children)` so `children=[]` is no longer treated as a group.

Both of your error/edge inputs are now shared golden cases: `group_without_logical_defaults_to_and` and `leaf_with_empty_children_is_leaf`, and each SDK's loader was taught the raw-group / empty-children shapes so they exercise the serializer's normalize path (not just the DSL).

README_GO_SDK.md
jymaire: Fixed in f939b6e — plumbed the error out rather than fixing the doc, as you suggested.

- `AppendFilterParams` / `appendFilterParams` now return `error` instead of panicking; `Where` no longer panics at construction.
- All ~70 hand-written search call sites and the 25 generated `*ByQuery` Execute sites propagate it through their existing `error` return (`return nil/"", err`). No panic is reachable from any public search or `*ByQuery` method on input-derived filters.
- The DSL constructors (`Eq`, `And`, `Or`, `Where`, …) keep returning `[]SearchFilter` so they stay composable — validation happens at serialization/request time.
- The Go README wording is updated to describe the error behavior (no panic).

Since the DSL is new in this PR, the `AppendFilterParams` signature change breaks no released consumer; noted in the PR description's signature-change section.

jymaire:
Thanks for the thorough end-to-end review, @Malaydewangan09 — built and ran all three suites, probed edges, checked against the UI encoder + backend binder. Everything is addressed in `f939b6e` (the 5 inline threads are replied + resolved; here are the review-body items):

**Blocking — Python second serializer.** Fixed. Both `parameters_to_tuples` and `parameters_to_url_query` in `kestrapy/api_client.py` now delegate to the shared `append_filter_params` (the url-query path re-`quote`s the values, preserving prior encoding); the flat-only copies that raised `ApiValueError` on a group node are gone. So the public `ApiClient` surface handles grouped filters like everything else.

**Nits:**
- **`regex` root collision** — done. The DSL helpers are no longer re-exported from the `kestrapy` root; import from `kestrapy.query`. Noted as a signature change in the description.
- **Java escaped-path untested** — added `escapedPathEncodesValues` (asserts `%20`/`%2C` on the `escape=true` path production uses), plus DSL drop/flatten/empty-group tests.
- **DSL drop/flatten + divergent empty-group predicate** — aligned. Java `isEmptyGroup` and Python `_is_empty_group` now use the same field-based rule, and both SDKs have tests for drop-empty-child / flatten-single-child / empty-group.
- **Go DSL only builds `[]SearchFilter`, the 25 `[]QueryFilter` endpoints got no DSL** — the `*ByQuery` path is now first-class: `[]QueryFilter` is converted and routed through the same recursive serializer (see the resolved thread on `legacy_client.go`), so grouped filters reach the wire and same-field OR duplicates survive.
- **READMEs implying the server refuses deeper nesting** — clarified in all three: the one-level cap is a client-side SDK choice; the server default is `maxDepth 3` / `maxWidth 20` (UI caps at 2).
- **`params.Set` → `params.Add`** — called out in the description's signature-change section.

**Unified group rule** (from the `query_filter.py` thread): a node is a group iff `logical != null` **or** it has a non-empty children list; null logical defaults to AND; empty children on a leaf is ignored; field+group data is ambiguous → client-side error. Both your error/edge inputs are now shared golden cases.

Offline suites after the change: Go 27, Java 24, Python 25 — all green.

Issue #14, opened 2025-07-03
Allow to create and update a flow from JSON (using the JSON builder instead of a YAML string)


AcevedoR:
- [ ] allow creating a Flow also from JSON rather than only from a YAML string (this would need to use a JSON builder)
- [ ] add tests for this use case

Very long issue as it needs to generate an entire model for every object from YAML. Parked until we hear a customer request for that (could be helpful e.g. for users who want to define an entire flow as a Pydantic class)

anna-geller:
can you explain a bit more what exactly do you mean? do you want to create a flow based on a local file path? 

Keep in mind that the SDK is autogenerated; it can only do what our API can do

Ben8t:
> allow creating a Flow easily -> not having to input Yaml

It's also about allowing the json builder to build a flow. The API already support that, the issue is coming from the SDK here.

jymaire:
## Implementation Plan

### Findings that change the framing

Two premises in this issue no longer hold (verified against `kestra` @ current `develop` and the `kestra-ee.yml` spec vendored in this repo):

1. **The API does *not* already accept JSON for flow create/update.** All three flow-write endpoints in `FlowController.java` are YAML-only:
   - `createFlow` — `@Post(consumes = { APPLICATION_X_YAML, APPLICATION_YAML })` (`FlowController.java:365`)
   - `updateFlow` — `@Put(uri = "{namespace}/{id}", consumes = { APPLICATION_X_YAML, APPLICATION_YAML })` (`:562`)
   - `updateFlowsInNamespace` / `bulkUpdateFlows` — same (`:426`, `:609`)

   Each takes a raw `@Body String` and parses it through `GenericFlow`. `kestra-ee.yml` agrees: `requestBody.content` for `createFlow` is `application/x-yaml` only. So the `area/backend` label is correct as the issue is literally worded — but see below, because there is a client-side route that needs no backend change.

2. **"Generate an entire model for every object" is not required.** The spec already settled this: `Task` (`kestra-ee.yml:42791`) is an *open envelope* — `id`, `type` (plugin class name), plus common fields (`retry`, `timeout`, `description`, …). Plugin-specific properties are deliberately unmodelled, and `Flow.tasks` carries `additionalProperties: true`. A `Flow` model already exists and ships in all four SDKs. The "very long issue" concern applies only to fully-typed per-plugin task classes, which is a different (and much larger) ask than what this issue needs.

### Design

[...]

PR #459, opened 2026-09-22
feat: create and update a flow from a native object (#14)

jymaire:
## What

Let callers build/update a Kestra flow from a native object (a typed `Flow` model **or** a plain dict/map) instead of hand-writing a YAML source string. Implemented in all four SDKs.

`closes: https://github.com/kestra-io/client-sdk/issues/14`

## Correction to the issue premise

An earlier comment on the issue suggested the API already accepts JSON for flow writes. That is **not** the case: the flow-write endpoints (`createFlow` POST, `updateFlow` PUT `{namespace}/{id}`, and the namespace/bulk variants) all declare `consumes = application/x-yaml` and do **not** accept JSON. So this is implemented **client-side**: the object is serialized to a YAML source string and posted to the existing YAML endpoints. **No backend change**, no new endpoint.

## How

Each SDK gains a serialization layer (object → YAML) and thin `*FromObject` methods that serialize then delegate to the existing string-body `create`/`update` methods (HTTP logic reused, not duplicated).

Serialization guarantees (all SDKs):
- block style (never inline/flow), no anchors/aliases (shared refs expanded, never `&id`/`*id`);
- `null` fields omitted rather than emitted;
- **server-managed / read-only fields stripped** (`deleted`, `revision`, `draft`, `tenantId`, `source`, `updated`) so the YAML mirrors what a user writes as flow source;
- key/insertion order preserved where the language allows (Go: for a typed `*Flow`; Go map inputs are alphabetically ordered — see below);
- Kestra expressions like `{{ inputs.foo }}` quoted, not mangled;
- multi-line strings (shell `commands`, python `script`) round-trip intact — emitted as **literal block scalars** in Python/Go/JS, and as **quoted scalars** in Java (jackson-dataformat-yaml only produces literal blocks with `MINIMIZE_QUOTES` on, which would unquote `{{ }}` expressions, so it is deliberately kept off);
- number precision preserved (Go decodes with `UseNumber()`, so a large `int64` is not coerced to `float64`);
- non-primitive values inside a dict input (nested models, enums, datetimes) converted at every depth (Python dict path no longer raises `RepresenterError`);
- `labels` as a list of key/value pairs **or** a map both serialize;
- non-ASCII kept verbatim (never escaped to `\uXXXX`);
- plugin-specific task properties preserved at **every** nesting depth (`tasks`, `errors`, `finally`, `afterExecution`);
- `draft` stays a query parameter, not a body field.

Go key-order note: a typed `*Flow` marshals struct fields in declaration order and that order is preserved; a Go `map` input cannot preserve insertion order because `encoding/json` emits map keys sorted, so map inputs are emitted with alphabetically ordered keys (documented on the method).

The object param accepts **both** the typed `Flow` model and a plain dict/map (the dict path is the ergonomic win the issue asks for).

## Java `Task` additional-properties bug fix (latent bug)

`io.kestra.sdk.model.Task` had **no** `@JsonAnyGetter`/`@JsonAnySetter`, so it **silently dropped every plugin-specific task property** (e.g. `message` on a Log task, `commands` on a Shell task) on serialization. This PR adds a `Map<String, Object>` additional-properties field with `@JsonAnyGetter`/`@JsonAnySetter` (and `equals`/`hashCode`/`toString` treatment), matching the openapi-generator additionalProperties idiom already used by the Python and Go `Task` models. Without this the Java `*FromObject` methods could not carry plugin task properties.

## Dependencies added

- Python: `pyyaml` (`pyproject.toml`)
- Go: `gopkg.in/yaml.v3` promoted from indirect to direct (`go.mod`)
- Java: `jackson-dataformat-yaml` (`build.gradle`)
- JS: `yaml` (`package.json`)

## Tests

Per-SDK **pure serialization unit tests** (no live server): build a flow with a plugin-specific property on a top-level task and inside a nested Sequential task, serialize, and assert the plugin properties are present + round-trip to the real values (multi-line strings, non-ASCII, null omission). Additional assertions cover the review findings: server-managed fields are stripped (all SDKs), no anchors/aliases for shared refs (Python), nested non-primitives serialize without error (Python), and large `int64` precision (Go).

Ran locally (builds/unit tests only — no Kestra booted, live integration suites left to the maintainer):
- **Python**: `python3 -m compileall` + import check OK; `pytest tests/test_flow_yaml.py` → **11 passed**.
- **Go**: `go build ./...` + `go vet ./...` clean; `go test -run TestFlowToYAML` → **8 passed**.
- **Java**: `./gradlew compileJava compileTestJava test --tests io.kestra.sdk.api.FlowYamlTest` → **7 passed**. (JUnit 5 added to `java-sdk/build.gradle` for the new source-level unit tests; the unused JUnit 4 dep was removed.)
- **JS**: `tsc --noEmit` on the SDK source → **clean (0 errors)**; the new serialization spec → **6 passed** (run with the SDK aliased to source, no live server). Note: the repo-level `npm run check:types` type-checks the *test* workspace against the built `dist/` and currently has **pre-existing, unrelated** failures (e.g. `AppsApi`/`AssetsApi` drift) plus a stale-`dist` miss of the new `flowToYaml` export until the SDK is rebuilt; my code itself type-checks clean against the generated types. No generator sources were changed, so `generate-sdks.sh` was not run.

### ⚠️ Signature change

**No existing method signatures changed** — every addition is a **new** method. All four SDKs expose a `draft` query parameter (never a body field). New public API added:
- Python: `FlowsApi.create_flow_from_object(tenant, flow, draft=None)`, `FlowsApi.update_flow_from_object(namespace, id, tenant, flow, draft=None)`
- Go: `FlowsAPI.CreateFlowFromObject(ctx, tenant, flow, draft *bool)`, `FlowsAPI.UpdateFlowFromObject(ctx, namespace, id, tenant, flow, draft *bool)`
- Java: `FlowsApi.createFlowFromObject(tenant, flow, draft)`, `FlowsApi.updateFlowFromObject(namespace, id, tenant, flow, draft)`
- JS: `createFlowFromObject({ flow, tenant?, draft? })`, `updateFlowFromObject({ flow, namespace, id, tenant?, draft? })`, and helper `flowToYaml(flow)`

Java `Task` gained a new `additionalProperties` accessor pair (`getAdditionalProperties`/`putAdditionalProperty`/`getAdditionalProperty`); this is additive and fixes the silent-drop bug described above.


python/python-sdk/kestrapy/base_api.py
```diff
@@ -354,5 +360,17 @@ def _construct_model(data: dict, model_type: Type[T]) -> T:
...
+            processed['additional_properties'] = extras
         return model_type.model_construct(**processed)
+
+    @staticmethod
+    def _validate_scalar(value: Any, field_type: Any) -> Any:
+        # Best-effort coercion of a scalar (e.g. an ISO timestamp string for a
+        # datetime field) so a constructed model does not keep the raw JSON
+        # value; the raw value is kept if it does not validate.
+        try:
+            return TypeAdapter(field_type).validate_python(value)
```
Malaydewangan09: A fresh `TypeAdapter` is built on every call. Measured on this branch:

```
uncached: 22.6 us/call
cached:    0.2 us/call
ratio:     119x
```

This sits on the shared `_construct_model` fallback, which `_deserialize` hits for every Python API, and `_construct_model` recurses into nested models and list items. A page of 100 objects with ~15 scalar fields each is ~1500 calls, so ~34ms of pure schema building per response. It's worse for a field whose annotation pydantic can't build a schema for: you pay the construction *and* the raised exception, every time, forever.

One-line fix:

```python
@staticmethod
@lru_cache(maxsize=None)
def _adapter(field_type):
    return TypeAdapter(field_type)
```

Separately, `except Exception` swallows everything, so a systematically broken annotation degrades to raw values with no signal. Worth narrowing to `ValidationError` at least.

java/java-sdk/src/main/java/io/kestra/sdk/api/FlowsApi.java
```diff
@@ -41,6 +52,43 @@ public class FlowsApi extends BaseApi {
     private static final String MULTIPART = "multipart/form-data";
     private static final String TEXT_CSV = "text/csv";
 
+    // Serializes a flow object to a YAML source string. The flow-write endpoints
+    // are YAML-only (they do not accept JSON), so a native object is serialized
+    // here and posted to the existing YAML endpoints. Configuration: block style,
+    // no document-start marker, no line wrapping, null fields omitted. Expressions
+    // like `{{ inputs.foo }}` are quoted (MINIMIZE_QUOTES stays disabled) and
+    // non-ASCII stays verbatim. LITERAL_BLOCK_STYLE is intentionally NOT enabled:
+    // jackson-dataformat-yaml only honors it when MINIMIZE_QUOTES is on, which
```
Malaydewangan09: The justification here isn't right. I tested it on jackson-dataformat-yaml 2.18.6 with both `MINIMIZE_QUOTES` and `LITERAL_BLOCK_STYLE` on:

```yaml
id: p
tasks:
- id: t
  expr: "{{ inputs.foo }}"    <- still quoted
  message: |                  <- literal block
    line1
    line2
  amb_yes: "yes"              <- still quoted
  amb_num: 1_000              <- plain
  amb_e: 1e3                  <- plain
```

`{{ }}` expressions stay quoted (SnakeYAML won't emit a scalar starting with `{` plain) and so do the YAML 1.1 booleans. The only gap is the number-like strings, which is exactly what the other three SDKs' `isAmbiguous*` predicate already handles, so a small `StdSerializer<String>` closes it.

As it stands Java emits this for the same input:

```yaml
id: "p"
namespace: "ns"
tasks:
- id: "t"
  message: "line1\nline2\n"
```

The server stores that verbatim as the flow source the user sees in the UI. Given the PR reorders Go's keys canonically for exactly that reason, Java being the odd one out here is worth another look. Current behaviour is safe, just not consistent.

python/python-sdk/kestrapy/flow_yaml.py
```diff
@@ -0,0 +1,166 @@
...
+    data = _make_serializable(data)
+    if isinstance(data, dict):
+        data = {k: v for k, v in data.items() if k not in _SERVER_MANAGED_FIELDS}
+    return data
+
+
+def flow_to_yaml(flow: Union[Dict[str, Any], Any]) -> str:
+    """Serialize a flow object (Flow model or dict) to a YAML source string."""
+    data = _to_serializable(flow)
+    return yaml.dump(
```
Malaydewangan09: No `width`, so PyYAML folds at 80 columns. Same for JS (`yaml` defaults `lineWidth: 80`). Go and Java don't fold. Probed all four with a long description and a long expression:

```yaml
# Python
description: a very long single line description that goes on and on and definitely
  exceeds eighty characters in total width

# JS  (note the \ continuation inside the quoted scalar)
expr: "{{
  xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx\
  xxxxxxxxxxxxxxxxx }}"
```

Both re-parse to the exact original string, so it's not a correctness bug. But that is the flow source the server stores and the UI shows, and it's the same readability argument the PR makes for canonical key order. `yaml.dump(..., width=4096)` and `{ lineWidth: 0 }` are one-liners each.

go-sdk/kestra_api_client/flow_yaml.go
```diff
@@ -0,0 +1,305 @@
...
+	"y": true, "yes": true, "n": true, "no": true,
+	"true": true, "false": true, "on": true, "off": true,
+}
+
+// ambiguousYAMLScalar matches strings that a YAML 1.1 / 1.2 or Jackson reader
+// may resolve to a number or timestamp: signed ints/floats with underscores
+// and exponents (with or without a dot or exponent sign), hex/octal/binary,
+// .inf/.nan, sexagesimal (12:30) and dates. Deliberately permissive: quoting a
+// string that did not need it is harmless, leaving one plain is not.
+var ambiguousYAMLScalar = regexp.MustCompile(`(?i)^[-+]?(` +
```
Malaydewangan09: This regex and the 26-string test list are copy-pasted verbatim into four places (`flow_yaml.go`, `flow_yaml.py`, `flow-from-object.ts`, and three test files) with nothing pinning them together. Four hand-written copies of the same contract will drift the first time someone adds a case to one of them.

A shared fixture listing `(input string -> must be quoted)` that all four assert against would be cheap here, and the repo already has that pattern under `test-utils/`.

Related: Java has no ambiguous-string test at all. It doesn't need the predicate today because it quotes everything, but if anyone acts on the `MINIMIZE_QUOTES` comment in `FlowsApi.java`, nothing catches the regression.

Malaydewangan09 (commented):
Built and ran every suite, probed the edge cases, and checked the YAML contract against Kestra core's actual reader (`JacksonMapper.ofYaml()` in `core/.../serializers/JacksonMapper.java`, which builds an `ObjectMapper` on `YAMLFactory`).

**The core premise checks out.** I fed the ambiguous-string list back through Jackson's YAMLMapper, the reader Kestra actually uses:

```
yes    -> BOOLEAN true      1_000 -> NUMBER 1000
off    -> BOOLEAN false     1e3   -> NUMBER 1000.0
true   -> BOOLEAN true      0755  -> NUMBER 83      (octal)
.inf   -> JsonParseException: Malformed numeric value '.inf'
"yes"  -> STRING            "1e3" -> STRING
```

Every one is covered by the predicate. `12:30`, `2026-09-23` and `y` come back as strings under Jackson, so those three are over-quoted, harmlessly. The regex is well calibrated, not guesswork. The `OutputValue` and `AbstractTrigger.when` fixes are real pre-existing bugs that only surface if you actually round-trip, good catches.

Suites I ran: Java `FlowYamlTest` 13 pass, Python `tests/` 46 pass, Go all pass.

Nothing blocks. Four things inline, plus nits:

**Nits:**
- `OutputValue` is an interface with no accessor, so reading a value means `((OutputValue.Value) output.getValue()).getValue()`. Declaring `Object getValue()` on the interface would drop the cast.
- `setup.py`'s `REQUIRES` didn't get `pyyaml`. Harmless in practice (I built the wheel: `[project].dependencies` wins, `Requires-Dist: pyyaml>=6.0` is in METADATA, setuptools even logs `install_requires overwritten in pyproject.toml`), but the two lists are now out of sync.
- Go leaves YAML-ambiguous map *keys* plain (`on: v1`, `yes: v2`) where Python/JS/Java quote them. I checked and this is harmless: Jackson's YAMLParser always reads mapping keys as strings, so Kestra never re-types a key. Mentioning it only because the other three do quote.
- JS `/all`'s default export is now `{ ...generated, ...flowFromObject }`. Property access is unchanged as you say, but the spread eagerly evaluates the whole generated namespace, and a future name collision would silently let `flow-from-object` win in the default export while ESM `export *` drops the ambiguous named export with no error.

python/python-sdk/kestrapy/flow_yaml.py
jymaire: Fixed in 1c23a74b: `yaml.dump(..., width=float("inf"))` in Python and `doc.toString({ lineWidth: 0 })` in JS. Each SDK gets a test with a long description and a 120-char expression that asserts a single unfolded line and an exact round-trip (`test_long_lines_are_not_folded` / `never folds long lines`). Both fail with the option removed.

java/java-sdk/src/main/java/io/kestra/sdk/api/FlowsApi.java
jymaire: You're right, the comment was wrong. Fixed in 1c23a74b: Java now enables `MINIMIZE_QUOTES` + `LITERAL_BLOCK_STYLE` and registers a `StringQuotingChecker` (via `YAMLFactory.builder().stringQuotingChecker(...)`) instead of a custom `StdSerializer<String>`. It extends `StringQuotingChecker.Default` and also quotes anything `isAmbiguousYamlString` flags, the same predicate as the other three SDKs. Because it hooks `needToQuoteName` too, keys are covered. Your example now emits:

```yaml
id: p
namespace: ns
tasks:
- id: t
  type: io.kestra.plugin.core.log.Log
  message: |
    line1
    line2
  expr: "{{ inputs.foo }}"
```

Covered by `multiLineStringsAreLiteralBlocksAndExpressionsStayQuoted` and the new fixture-driven `ambiguousStringsAreQuotedAndRoundTrip`, which re-reads through Jackson's YAMLMapper and asserts every value is textual. With the custom checker removed, that test fails on `1_000`.

go-sdk/kestra_api_client/flow_yaml.go
jymaire: Done in 1c23a74b. `test-utils/yaml-ambiguous-strings.json` now holds `mustQuote` (the 26 strings) and `staysPlain`. The Go, Python, JS and Java tests all load it instead of their own copies, and each asserts:
- every `mustQuote` string is quoted as a value **and** as a map key;
- every `staysPlain` string stays plain;
- the output re-parses to the exact strings (Jackson in Java, PyYAML 1.1 in Python, `yaml` 1.1+1.2 in JS, yaml.v3 in Go).

Java now has this test, and it guards the `MINIMIZE_QUOTES` switch from the other thread. The four source predicates now point at the fixture in their comments.

One thing I found: an empty-string *key* is left out of the key checks. SnakeYAML writes it as the complex key `? ""`, which is valid and reads back as `""`, just not in the `"x": k` shape.

python/python-sdk/kestrapy/base_api.py
jymaire: Fixed in 1c23a74b. `_validate_scalar` now goes through a module-level `lru_cache`d adapter per annotation, including the "pydantic can't build a schema" result (cached as `None`), so a bad annotation costs one failed construction instead of one per call. Annotations that can't be hashed fall back to an uncached build. The `except` is narrowed: `ValidationError` on validate, and `PydanticUserError`/`NameError`/`TypeError` on construction only. New `test_validate_scalar_coerces_and_caches_adapters` checks that three calls give 1 miss and 2 hits, plus the coerced datetime value.

jymaire:
Thanks @Malaydewangan09, thorough review. All four inline threads are fixed in 1c23a74b (replied on each). On the nits:

- **`OutputValue` accessor**: `Object getValue()` is now declared on `OutputValue`, `PropertyListAsset` and `PropertyListAssetIdentifier`, so `output.getValue().getValue()` needs no cast. It only matters for a third-party class implementing one of these interfaces. That's noted under ⚠️ Signature change in the description.
- **`setup.py` `REQUIRES`**: added `pyyaml >= 6.0`, so it matches `pyproject.toml` again.
- **Go map keys**: now quoted when ambiguous, like the other three. As you found, Jackson always reads keys as strings, so this is only for consistency. The shared fixture test now checks keys in all four SDKs.
- **JS `/all` default export**: on eager evaluation, `import * as generated` already evaluates the whole module whatever the default export is, so the spread only copies bindings that already exist. It doesn't pull in anything extra. The collision risk is real, though, so I added a test: every key of the default export must be the *same value* as the named export. A shared name fails it both under real ESM (named export dropped) and under vitest's transform (named export resolves to the generated one). I checked that it fails by adding a clashing `createFlow` export to `flow-from-object`.

Local runs: Java unit tests 41/41, Python `tests/` 48/48, Go `kestra_api_client` pass, JS `FlowFromObjectApi.spec.ts` 13/13 (offline), and `check:types` clean. I didn't run the live suites.

jymaire:
### Live QA re-run on `1c23a74b` (review fixes)

The review fixes change what goes over the wire (the Java YAML emitter, Go key quoting, Python/JS line folding), so I re-ran the full manual QA against a live `kestra-ee:develop-slim` (the image from 2026-09-22, the same one used for the first QA). All four SDKs pass.

**Writes:** for each SDK, a rich flow created and updated from a plain map, a typed round-trip (`flow()` → `update*FromObject`), a typed build with plugin properties on the task and trigger, and a create with `draft=true`. All 16 writes succeeded, with revision 1 on create and 2 on update, and `draft=true` respected.

**Executions + stored source.** Each flow was run, then its task outputs, logs and stored `source` were checked. The rich flow has inputs, variables, a nested `Sequential`, `errors`/`finally`, flow outputs, a Schedule trigger, map labels, multi-line / unicode / `{{ }}` strings, an int64, an empty list, and 26 YAML-ambiguous strings. The update also sets a label to `on`.

| Flow | Python | Go | Java | JS |
|---|---|---|---|---|
| Rich map (15 checks) | 15/15 | 15/15 | 15/15 | 14/15 ¹ |
| Typed round-trip (15 checks) | 15/15 | 15/15 | 15/15 | 14/15 ¹ |
| Typed build (3 checks) | 3/3 | 3/3 | 3/3 | 3/3 |

¹ `9007199254740993` → `…992`: JavaScript numbers can't hold integers above 2^53 exactly. This is the known gap already listed in the description.

A control flow sent as raw PyYAML `safe_dump` scores 14/15: `1e3`/`1E-3` come back as numbers. That shows the verifier catches strings the server re-types.

**The changes in this commit, checked in the stored source:**
- **Java:** multi-line strings are `|` blocks, and other strings are plain where safe. `{{ }}` expressions and ambiguous values (`"yes"`, `"on"`, …) are still quoted, and all 26 ambiguous strings round-trip unchanged. `MINIMIZE_QUOTES` does not unquote expressions.
- **Go:** ambiguous map keys are now quoted (`"on"`, `"off"`, `"Yes"`).
- **Python / JS:** a 228-character message containing `{{ flow.id }}` is stored on one line and is logged correctly after rendering.

Malaydewangan09 (approved):

Malaydewangan09 (commented):
Re-reviewed `1c23a74` end to end. All four items are fixed, and I verified each one rather than reading the diff.

**Cross-SDK output is now consistent.** Same input through all four serializers:

```yaml
                              # Go / Java / Python / JS, all four:
id: compare
namespace: company.team
description: a very long single line description that goes on and on and definitely exceeds eighty characters in total width
labels:
  approved: "yes"
  env: prod
  "on": v1
tasks:
  - id: t
    type: io.kestra.plugin.core.log.Log
    amb: ["yes", "1_000", "1e3", "0755", ".inf", "", "+1"]   # each on its own line, all quoted
    bignum: 9007199254740993
    emptylst: []
    expr: '{{ inputs.foo }}'      # Go/Python single, Java/JS double, same value
    message: hello world
    plain: my-flow
    script: |
      echo one
      echo two
    unicode: héllo wörld ✓
```

Byte-identical apart from sequence indentation and quote style. Java in particular went from `id: "p"` and an escaped `message` to the same literal blocks and plain scalars as the rest. No folding anywhere. `{{ }}`, ambiguous values, ambiguous keys, int64 and unicode all hold.

**The `except ValidationError` narrowing is safe.** That was the one change I thought could regress, since it used to swallow everything. I fuzzed `_construct_model` over **561 models x every field x 15 junk values = 47,925 combinations**. Nothing escaped `_validate_scalar`. The only `TypeError` I saw was my own fuzz artifact (a field literally named `cls` colliding with `model_construct`'s parameter, at the `model_construct(**processed)` call, unrelated and pre-existing).

Also checked: `_scalar_adapter`'s outer `except TypeError` correctly catches only the unhashable-annotation case, because `_build_scalar_adapter` already swallows a `TypeError` from `TypeAdapter(...)` itself. `PydanticSchemaGenerationError` extends `PydanticUserError` and `PydanticUndefinedAnnotation` extends `NameError`, so both are covered. Nice touch that the cache test asserts `(misses, hits) == (1, 2)`, pinning the caching and not just the coercion.

**The shared fixture is real.** All four read `test-utils/yaml-ambiguous-strings.json` from disk and fail loudly if it's missing (Python `FileNotFoundError`, Java `IllegalStateException`), so none of them can silently skip. Java got the test it was missing.

Suites I ran: Java `FlowYamlTest` 15 pass, Python `tests/` 48 pass, Go all pass. All four CI checks green.

One leftover, in the PR description rather than the code:

- The serialization-guarantees bullet still says multi-line strings are emitted *"as **quoted scalars** in Java (jackson-dataformat-yaml only produces literal blocks with `MINIMIZE_QUOTES` on, which would unquote `{{ }}` expressions, so it is deliberately kept off)"*, and the re-typing bullet still ends with *"(Java already quoted every string)"*. Both describe the old behaviour and now contradict the code.

Nothing else from my side.

PR #462, opened 2026-09-22
feat(sdk): guard replay_execution_with_inputs against an empty body (#461)

### LINKED ISSUE #461: SDK: guard replay_execution_with_inputs against an empty body (endpoint requires it)
## Summary

`POST /executions/{executionId}/actions/replay-with-inputs` declares its `multipart/form-data` request body as **`required: true`** (`kestra-ee.yml`). But every hand-written SDK exposes `inputs` as **optional** and, when it's omitted, sends a request with **no body**. The caller gets a cryptic server error instead of a clear client-side one:

```
422 {"title":"Invalid request body",
     "detail":"Required Body [inputs] not specified",
     "instance":"/api/v1/main/executions/{id}/actions/replay-with-inputs"}
```

This proposes a small **non-breaking** client-side guard so the SDK fails fast with an actionable message and points the caller at the right method.

## Why it happens

The endpoint always needs a body. The SDKs don't enforce that:

| SDK | Method | Behaviour with no `inputs` |
|-----|--------|-----------------------------|
| Python | `replay_execution_with_inputs` | `inputs: Optional[Dict] = None` → `_build_inputs_multipart(None)` returns `None` → no body → **422** |
| Go | `ReplayExecutionWithinputs` | `.FormData(...)` optional; `formData` defaults `nil` → no body → **422** |
| Java | `replayExecutionWithInputs` | `inputs @Nullable`; `toFormParams(null)` → empty map → sends a present (empty) body — no client-side check either way |
| JS | `replayExecutionWithinputs` (generated) | body optional in the generated call |

`replay_execution_with_inputs` with no inputs is not a meaningful operation — replaying **without** changing inputs is what `replay_execution` (plain `.../actions/replay`) is for.

## Scope (verified against the spec)

Among the execution endpoints, **only `replay-with-inputs` has a `required: true` multipart body**. `resume`, `resume/validate`, and `create_execution` (`/executions/{namespace}/{id}`) all have **optional** bodies, so they are unaffected — a flow with no inputs legitimately sends no body there. (`flows/validate` and `policies/import` are also `required: true`, but the SDK always sends their content, so there is no gap.)

## Proposal — non-breaking guard

Do **not** make `inputs` a required parameter (that's a breaking signature change and conflicts with the additive-changes convention). Instead, in each hand-written SDK, when `replay_execution_with_inputs` is called with empty/absent `inputs`, raise a clear client-side error **before** the HTTP call, e.g.:

> `replay_execution_with_inputs requires a non-empty 'inputs' body; to replay without changing inputs, use replay_execution instead.`

- **Python** — raise `ValueError` when `inputs` is falsy.
- **Go** — return an error from `Execute()` when `formData` is nil/empty (don't panic).
- **Java** — throw `ApiException` (or `IllegalArgumentException`) when `inputs` is null/empty.
- **JS** — mirror via the hey-api SDK wrapper plugin.

Keep the signature and the optional parameter as-is; only the empty-call path changes from "opaque 422" to "clear client-side error".

### Acceptance criteria
- [ ] Calling `replay_execution_with_inputs` with no/empty `inputs` raises a clear client-side error in Py, Go, Java, JS (no network round-trip, no panic).
- [ ] Calling it **with** inputs is unchanged and still round-trips (existing `..._overrides_inputs` coverage stays green).
- [ ] Method signatures unchanged (no breaking change); note in the PR per `AGENTS.md`.

## Context

Surfaced while investigating a red Python CI job. The immediate CI unblock — removing two tests that exercised the unsupported no-input call, plus a docstring fix — is **#460**. This issue is the follow-up hardening to give callers a good error across all four SDKs.

jymaire:
Closes #461.

## Problem

`POST /executions/{executionId}/actions/replay-with-inputs` declares its multipart request body as **`required: true`** (`kestra-ee.yml`). A call with no inputs is rejected server-side with an opaque error:

```
422 {"title":"Invalid request body","detail":"Required Body [inputs] not specified"}
```

All three hand-written SDKs expose `inputs` as optional and, when it's omitted, send a boundary-less body that the server rejects — so callers hit that cryptic 422 with no hint that they should have used `replay_execution` (a call with no inputs to override is not a supported operation on this endpoint).

## Change — non-breaking client-side guard

`inputs` stays optional in every signature (additive, no breaking change); only the empty-call path changes, from an opaque server 422 to a clear client-side error naming `replay_execution`:

| SDK | Guard | Error |
|-----|-------|-------|
| Python | `ExecutionsApi.replay_execution_with_inputs` raises when `inputs` is falsy | `ApiValueError` |
| Java | `ExecutionsApi.replayExecutionWithInputs` throws when `inputs` is null/empty | `ApiException` |
| Go | `ExecutionsAPI.ReplayExecutionWithInputs` (and the legacy `ExecutionsAPIService.Execute`) return an error when the inputs form is empty | `error` |

Each guard short-circuits **before any HTTP request**, and each SDK gets an **offline** guard test (no server needed): Python `tests/test_replay_with_inputs_guard.py`, Go `test/api_executions_replay_guard_test.go`, Java `ExecutionsApiReplayGuardTest`. All three pass locally.

## Live verification

Probed a Kestra 2.0 instance directly: an empty body with a boundary-less `multipart/form-data` content-type returns the `422`, while a *well-formed* empty multipart (valid boundary, 0 parts) **binds and succeeds**. All three SDKs currently send the boundary-less shape for empty inputs (Python `files=None`; Java `StringEntity("", "multipart/form-data")`; Go skips the multipart writer when the form is empty), so every one already gets the 422 today — the guard breaks no currently-working call. The wire shape is deliberately left as-is; the guard is client-side only.

### ⚠️ Signature change
None. `inputs` remains an optional parameter in all three SDKs.

## Note on #460

This also removes two degenerate no-input `replay_execution_with_inputs` tests from the Python integration suite (they only ever exercised the now-guarded unsupported call; the `TestReplayBulk` one additionally called the non-bulk method). #460 removed the same two tests as a minimal CI unblock — **this PR supersedes #460**, which can be closed once this merges.


jymaire:
Closing — this took the wrong direction.

Main-branch CI (Go `b2888718`, Java `5a67217b`, both green) shows that **Go and Java already support no-input `replay-with-inputs` and return 200**: their wrappers send a *well-formed* empty multipart body (Go's `doMultipartJSON` always closes the writer; Java has a deliberate 5-arg overload with no inputs), and the server binds it. Only **Python** fails (it sends no body at all → 422).

So a client-side guard that rejects empty inputs is a **breaking change** for Go and Java (it removed working, tested behaviour — that's why this PR's CI went red on the existing `replayExecutionWithInputsTest` / `replayExecutionWithInputs_basic` tests, not the RBAC baseline).

The real fix is Python-only: make it send a well-formed empty multipart like Go/Java so no-input replay works there too. See #461 for the corrected plan.

